PR #19, opened 2017-03-09
CC-500: Provide exactly-once time-based partitioning in S3

kkonstantine:


kkonstantine:
Includes the necessary changes to provide exactly-once semantics given deterministic, timestamp-based partitioners. Requires the partitioners defined in storage-common: https://github.com/confluentinc/kafka-connect-storage-common/pull/18

Increasing test coverage of time-based partitioning. More tests will be added soon for partitioning based on user-defined timestamp column. /cc @confluentinc/connect

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -102,8 +98,9 @@ public TopicPartitionWriter(TopicPartition tp,
 
     flushSize = connectorConfig.getInt(S3SinkConnectorConfig.FLUSH_SIZE_CONFIG);
     topicsDir = connectorConfig.getString(StorageCommonConfig.TOPICS_DIR_CONFIG);
-    rotateIntervalMs = connectorConfig.getLong(S3SinkConnectorConfig.ROTATE_INTERVAL_MS_CONFIG);
-    rotateScheduleIntervalMs = connectorConfig.getLong(S3SinkConnectorConfig.ROTATE_SCHEDULE_INTERVAL_MS_CONFIG);
+    rotateIntervalMs = partitioner instanceof TimeBasedPartitioner ?
```
kkonstantine: I was thinking that using a single interval (duration for that matter) to adjust the granularity of time buckets and then depend on file size for additional file rotation within a time bucket would be good enough. Giving it a second thought, I agree that treating time buckets as more static, and using a time-based configurable interval, such as `rotate.interval.ms`, besides flush size to rotate files within the bucket makes sense. Will allow for more fine-grain rotation. 

Having said that, this property doesn't solve the issue of outstanding messages. Since we don't use scheduled intervals (wallclock based) we always depend on the "next" message to perform rotation, file-based or time-based. For instance, with hourly time-buckets, and rotation interval of 20 minutes, if messages suddenly stop at, let's say, hh:58 then all records between hh:40 and hh:59 will remain outstanding until the "next record" arrives.

I feel that's a reasonable trade-off for exactly-once time-based partitioning.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -143,28 +133,13 @@ public State next() {
     }
   }
 
-  private void updateRotationTimers() {
-    long lastRotate = time.milliseconds();
-    if (log.isDebugEnabled() && rotateIntervalMs > 0) {
-      log.debug("Update last rotation timer. Next rotation for {} will be in {}ms", tp, rotateIntervalMs);
-    }
-    if (rotateScheduleIntervalMs > 0) {
-      nextScheduledRotate = DateTimeUtils.getNextTimeAdjustedByDay(lastRotate, rotateScheduleIntervalMs, timezone);
```
kkonstantine: Using `rotate.interval.ms` in addition to `partition.duration.ms` will require some code in the partitioner to compare timestamps between records. And to do that we'll have to adjust time as above. Currently we do that in [getPartition](https://github.com/kkonstantine/kafka-connect-storage-common/blob/170f848b7f244a3ffda8f509342a8caf4d94b1bb/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java#L73-L75)

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
ewencp: Yeah, I guess rotate interval can't really be used if you want exactly once. Some folks may be ok without that guarantee, especially if they can guarantee the latency of delivery.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
ewencp: Ok, makes sense. As mentioned on the other comment, we might want this functionality for those that aren't worried about exactly once, but fine to remove it as long as we're not supporting that.

kkonstantine: I hear you. We should give users the option to schedule rotations, making clear at the same time that they abolish exactly-once this way. 

I think reasonable defaults according to the discussion above are: 

`rotate.interval.ms` to default to `partition.duration.ms` for users that don't care about setting another interval (e.g. they want everything under hourly time-buckets where files are rolled based on size below that)
`schedule.interval.ms` to be off by default (`<0`) to aim for exactly-once by default unless you change it explicitly.

ewencp: Why wouldn't `rotate.interval.ms` default to `-1`, which will disable using wallclock time rotation entirely? Maybe part of the confusion here is that in the HDFS connector we only had the concept of wallclock time so anything time-based is wallclock time but the S3 connector uses the timestamp from the record? In which case the docs for `rotate.interval.ms` taken from HDFS don't make sense since the same implementation cannot accomplish the stated goal of `rotate.interval.ms`.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -255,17 +233,19 @@ private void setState(State state) {
...
-                                    && timestamp != null
-                                    && timestamp >= nextScheduledRotate;
-    boolean messageSizeRotation = recordCount >= flushSize;
-
-    log.trace("Should rotate (count {} >= flush size {} and schedule interval {} next schedule {} timestamp {})? {}",
-              recordCount, flushSize, rotateScheduleIntervalMs, nextScheduledRotate, timestamp,
-              scheduledRotation || messageSizeRotation);
+  private boolean rotateOnTime(String encodedPartition) {
+    boolean timeBasedRotation =
+        rotateIntervalMs > 0 && recordCount > 0 && !encodedPartition.equals(currentEncodedPartition);
```
schizhov: Messages are not guaranteed to arrive ordered with respect to the timestamp (especially with `CreateTime` timestamps), which might result in lots of small files being created around the partition boundary. Or am I missing something?

kkonstantine:
Thanks for your review @ewen! 

The last commit included a few adjustments in the time-based unit tests. This created: https://github.com/confluentinc/kafka-connect-storage-common/pull/23 which needs to be merged first before this PR builds successfully. 

Also, I think this (as well as the above PR) should be rebased on top of 3.2.x. @ewen let me know if you agree.

kkonstantine:
ok to test

kkonstantine:
@ewencp thanks so much for the reviews!

PR #20, opened 2017-03-15
involve rotate.interval.ms in rotation check

dmnpignaud:
The "rotate.interval.ms" parameter wasn't used in S3 file rotation check. 
I just copied the behavior implemented in qubole/streamx : https://github.com/qubole/streamx/blob/master/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -77,6 +77,7 @@ public class TopicPartitionWriter {
   private final String fileDelim;
   private final DateTimeZone timezone;
   private final Time time;
+  private final long lastRotate;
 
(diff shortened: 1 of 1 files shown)


kkonstantine:
Hi @dmnpignaud, 

thanks for submitting a fix. Enhancing time-based partitioning in the S3 connector is already underway, here: https://github.com/confluentinc/kafka-connect-storage-common/pull/18 and here: https://github.com/confluentinc/kafka-connect-storage-cloud/pull/19) 

There are a few more changes to be made besides just checking against wall-clock time. However, let me note that, although we will allow this to be configurable, by taking into account the wall-clock time in file rotation, someone will prevent the connector from guaranteeing exactly-once semantics.

kkonstantine:
The suggested change has been incorporated after https://github.com/confluentinc/kafka-connect-storage-cloud/pull/19
Given that, I'd suggest closing this PR. Thanks for taking a stab at this @dmnpignaud !

PR #25, opened 2017-03-29
flush.size doc fixes

cotedm:
We should mention this parameter is per partition. @kkonstantine can you double check me? Should be quick.


kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -147,7 +147,7 @@ The example settings are contained in ``etc/kafka-connect-s3/quickstart-s3.prope
   flush.size=3
 
 The first few settings are common to most connectors. ``topics`` specifies the topics we want to export data from, in
-this case ``s3_topic``. The property ``flush.size`` specifies the number of records the connector needs to write before
+this case ``s3_topic``. The property ``flush.size`` specifies the number of records per partition the connector needs to write before
```
kkonstantine: nit: Would you mind adding a line break around the column width of this text (here 120 chars) so the text won't wrap for a single line? (and then merge with the line below).

PR #30, opened 2017-04-07
S3 Proxy Settings

vasanthrj:


ewencp:
ok to test

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -32,8 +32,7 @@
 import io.confluent.connect.storage.Storage;
 import io.confluent.connect.storage.common.util.StringUtils;
 
-import static io.confluent.connect.s3.S3SinkConnectorConfig.WAN_MODE_CONFIG;
-import static io.confluent.connect.s3.S3SinkConnectorConfig.REGION_CONFIG;
+import static io.confluent.connect.s3.S3SinkConnectorConfig.*;
```
ewencp: nit: We try to avoid * imports since they clutter the local namespace in the file.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -139,6 +142,24 @@
                         ++orderInGroup,
                         Width.LONG,
                         "S3 accelerated endpoint enabled");
+
+      CONFIG_DEF.define(S3_PROXY_HOST,
+              Type.STRING,
+              Importance.HIGH,
```
ewencp: I'd say these are probably medium or even low importance -- many (most?) users won't even use these settings, so high is probably overkill.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -139,6 +142,24 @@
...
+              Width.LONG,
+              "proxy host");
+
+      CONFIG_DEF.define(S3_PROXY_PORT,
+              Type.INT,
+              Importance.HIGH,
+              "proxy port.",
+              group,
+              ++orderInGroup,
+              Width.LONG,
```
ewencp: Width can probably be SHORT here -- ports should only be a few digits long at most.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -60,12 +59,14 @@ public S3Storage(S3SinkConnectorConfig conf, String url) {
...
   public AmazonS3 newS3Client(S3SinkConnectorConfig config) {
+    String proxyHost = config.getString(S3_PROXY_HOST);
+    int proxyPort = config.getInt(S3_PROXY_PORT);
     AmazonS3ClientBuilder builder = AmazonS3ClientBuilder.standard()
                                         .withAccelerateModeEnabled(config.getBoolean(WAN_MODE_CONFIG))
                                         .withPathStyleAccessEnabled(true)
                                         .withCredentials(config.getCredentialsProvider())
                                         .withClientConfiguration(
-                                            PredefinedClientConfigurations.defaultConfig()
+                                            PredefinedClientConfigurations.defaultConfig().withProxyHost(proxyHost).withProxyPort(proxyPort)
```
ewencp: I think you need to make this conditional on whether the values are actually set or not -- not all users will be using a proxy, so we can't always call these methods.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -139,6 +142,24 @@
                         ++orderInGroup,
                         Width.LONG,
                         "S3 accelerated endpoint enabled");
+
+      CONFIG_DEF.define(S3_PROXY_HOST,
```
ewencp: The override you're calling here is for a required field. Since not all users will need proxy settings, you should call an override of `define` that includes default values. These can just be placeholders (i.e. `""` and -1).

The current CI test failures are because all the existing tests do not expect to need to set these configs, but the config parsing checks that all required fields are properly filled in.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -139,6 +142,24 @@
...
+              Type.STRING,
+              Importance.HIGH,
+              "proxy host.",
```
ewencp: These should be straightforward to understand, but perhaps we could get a full sentence description for the documentation?

nit: Fixing the capitalization would be nice too :)

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -139,6 +142,24 @@
...
+              group,
+              ++orderInGroup,
+              Width.LONG,
+              "proxy host");
```
ewencp: nit: Could we match the capitalization/style of the other display names?

vasanthrj:
@ewencp Implemented review comments. please review again.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -63,6 +63,11 @@
   public static final String REGION_CONFIG = "s3.region";
   public static final String REGION_DEFAULT = Regions.DEFAULT_REGION.getName();
 
+  public static final String S3_PROXY_HOST = "s3.proxy.host";
+  public static final String S3_PROXY_PORT = "s3.proxy.port";
+  public static final String S3_PROXY_HOST_DEFAULT = null;
+  public static final int S3_PROXY_PORT_DEFAULT = 0;
```
kkonstantine: Should be `-1` to match the default in AWS sdk: https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java#L155

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -63,6 +63,11 @@
   public static final String REGION_CONFIG = "s3.region";
   public static final String REGION_DEFAULT = Regions.DEFAULT_REGION.getName();
 
+  public static final String S3_PROXY_HOST = "s3.proxy.host";
```
kkonstantine: I'm a bit torn between having two properties, or just adding `S3_PROXY` and (eventually?) support all the properties in one string, as described here: http://docs.aws.amazon.com/cli/latest/userguide/cli-http-proxy.html (the rest are protocol, username and password, and a couple more on windows ...)

In that case though, it seems that protocol should be supported from the start (or maybe just accept `host` or `host:port` for now, and later add support for `protocol://username:pass@host:port` and valid combinations)

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -60,14 +62,20 @@ public S3Storage(S3SinkConnectorConfig conf, String url) {
   }
 
   public AmazonS3 newS3Client(S3SinkConnectorConfig config) {
+    String proxyHost = config.getString(S3_PROXY_HOST);
+    int proxyPort = config.getInt(S3_PROXY_PORT);
+
+    ClientConfiguration clientConfiguration = PredefinedClientConfigurations.defaultConfig();
+    clientConfiguration.withUserAgentPrefix(String.format(VERSION_FORMAT, Version.getVersion()));
+    if(proxyHost != null && !proxyHost.isEmpty()) {
```
kkonstantine: Also, here regarding the previous comment from @ewencp : 
the previous use of withers had the advantage that, if the defaults (`null` and `-1`) were set with the "withers", aws sdk would also check for the system properties `https.proxyHost` and `https.proxyPort`. @ewencp is this you think an option we want to have along with our configuration properties? Or we should override it with our configuration regardless?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
vasanthrj: But AWS API ClientConfiguration accepts hostname and port separately. If we accept as single string then it requires splitting.

kkonstantine: Yes, sorry if it wasn't clear, I implied that splitting/parsing of a single string would be required in order to extract the arguments.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
vasanthrj: System properties https.proxyHost and https.proxyport taken care in ClientConfiguration.java.
https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java#L602

kkonstantine: Right, that's where I'm getting at. If you do `withProxyHost(null)`, aws sdk will eventually check the appropriate system property when it'll use this config (for instance `https.proxyHost`). The check we add here for proxyHost being not null is kind of a no-op.

Here's the code that is used internally to determine whether a proxy will be used: 
https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/http/apache/client/impl/ApacheHttpClientFactory.java#L116 

aws sdk will check the system properties anyways, even if you don't call `setProxyHost` or `withProxyHost`. That also means that setting a proxy is possible even with the current config (although I'm totally up for adding this configuration in S3 Connector's properties as well).

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
ewencp: @kkonstantine Using just a URL has its own problems, e.g. that the password still needs to be added as a separate field to avoid leaking secrets in log files (or you just need to make the entire URL a `PASSWORD` field).

I don't know much about how people use proxies for S3 in practice, so I'm not sure which features listed on that page are important. For example, besides host and port, I could imagine that SSL might be required for something like this.

kkonstantine:
Hi @vasanthrj, it's been a while since we discussed this one. I'd be happy to pick it up and add the few fixes that are left to tackle, if that's ok with you. I could try to push the changes to your branch or issue a new PR if that's not possible. Let me know what do you think.

vasanthrj:
@kkonstantine Addressed review comments. If you have more fixes then take my changes to your branch and create your own PR that would be better. I can close this PR.

kkonstantine:
This PR was cloned to and superseded by https://github.com/confluentinc/kafka-connect-storage-cloud/pull/97 
Thus, I'm closing without merging. Thanks @vasanthrj for taking the initial stab on this one.

PR #33, opened 2017-04-14
CC-513: Newline is not appended between records when using JsonFormat

kkonstantine:
Jackson as default root value separate a single white space.

Changing it to the line separator used by the system running the connector. (`"\n"` for Unix systems, `"\r\n"` for Windows systems).

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
@@ -17,6 +17,7 @@
 package io.confluent.connect.s3.format.json;
 
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.io.SerializedString;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.kafka.connect.data.Struct;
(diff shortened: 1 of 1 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
```diff
@@ -58,7 +60,9 @@ public RecordWriter getRecordWriter(final S3SinkConnectorConfig conf, final Stri
     try {
       return new RecordWriter() {
         final S3OutputStream s3out = storage.create(filename, true);
-        final JsonGenerator writer = mapper.getFactory().createGenerator(s3out);
+        final JsonGenerator writer = mapper.getFactory()
+                                         .createGenerator(s3out)
+                                         .setRootValueSeparator(new SerializedString(LINE_SEPARATOR));
```
ewencp: Are we ok if the last line in the file doesn't have a newline character? Not sure if anything would be affected by that, but based on the behavior we saw during debugging, presumably with this code the last line would be missing it.

kkonstantine: Right. Let's not. Let's add it ourselves.

PR #34, opened 2017-04-14
CC-524: Reduce the default s3.part.size to 25MB to avoid OOM with current...

kkonstantine:
… default jvm heap size


gwenshap:
I'd prefer the smallest possible size (optimizing for less memory and lower throughput), but I know you guys had different thoughts on user preferences. Let's go with this size and see if the user complaints go away?

PR #37, opened 2017-04-20
Enable compression of output Avro files 

niksajakovljevic:
Add support for Avro compression codec. This will enable Avro file compression with codecs provided by Avro library.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -63,6 +63,9 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
   public static final String REGION_CONFIG = "s3.region";
   public static final String REGION_DEFAULT = Regions.DEFAULT_REGION.getName();
 
+  public static final String AVRO_CODEC_CONFIG = "avro.codec";
+  public static final String AVRO_CODEC_DEFAULT = "deflate";
+
   private final String name;
 
   private final StorageCommonConfig commonConfig;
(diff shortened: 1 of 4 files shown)


ewencp:
ok to test

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -63,6 +63,9 @@
   public static final String REGION_CONFIG = "s3.region";
   public static final String REGION_DEFAULT = Regions.DEFAULT_REGION.getName();
 
+  public static final String AVRO_CODEC_CONFIG = "avro.codec";
+  public static final String AVRO_CODEC_DEFAULT = "deflate";
```
ewencp: Setting this default would change the existing behavior of the connector (i.e. an existing user would suddenly be getting deflated files instead of uncompressed). I'm also not sure deflate is a default most would want these days (e.g. is the decision based on performance/compression tradeoff? splittability? something else?).

niksajakovljevic: Good point. I am thinking to set it to `null` by default as it will not affect existing behavior.

niksajakovljevic:
@ConfluentCLABot [clabot:check]

niksajakovljevic:
Hi @ewencp , I've added test for Avro file compression. Please have a look. I've also changed default codec to be `null`. I was trying to scope changes for this pull request only to Avro compression so I think it's better to handle GZip and `.gz` naming in the separate pull request. I assume that support for GZip should be done with the new format.

ewencp:
Sorry, Gzip and `.gz` was just an example; I really just meant whether the different formats should have different extensions to indicate what is stored in the file. But I guess it doesn't really matter since this is all internal to the Avro Object Container file format anyway.

PR #41, opened 2017-05-04
Fix formatting of the changelog.

ewencp:


kkonstantine:
It's strange that rst doesn't understand that there's no real break when you wrap a line ... which is a pity for those who have auto-wrap enabled.

PR #45, opened 2017-05-08
HOTFIX: S3SinkConnector should extend SinkConnector

kkonstantine:
* Needed by the framework for better management of the connector.
  * Fixes CC-537: S3SinkConnector needs to extend SinkConnector for runtime class resolution


norwood:
😍

PR #60, opened 2017-06-27
Add logging when creating a new RecordWriter

ewencp:


ewencp:
Simple logging enhancement that will make it easier to track down the source of things like memory use issues from the logs.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -349,6 +349,11 @@ private RecordWriter getWriter(SinkRecord record, String encodedPartition) throw
       return writers.get(encodedPartition);
     }
     String commitFilename = getCommitFilename(encodedPartition);
+    log.debug(
+        "Creating new writer encodedPartition={} filename={}",
```
kkonstantine: In similar log lines I've found that is useful to enclose the values in single quotes `'`. Makes reading of such lines easier when the value is empty. Low priority.

PR #72, opened 2017-07-19
CC-967: Fix committing offset when rotating on time or schema.

kkonstantine:


kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -413,7 +408,7 @@ private void commitFiles() {
       commitFile(entry.getKey());
       log.debug("Committed {} for {}", entry.getValue(), tp);
     }
-    offsetToCommit = nextOffsetToCommit;
+    offsetToCommit = currentOffset + 1;
```
rhauch: Do you add 1 because this offset gets committed to offset storage, and upon restart that's where the connector will start reading?

kkonstantine: Right, that's the convention.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -63,17 +66,21 @@ public S3SinkTask() {
     // no-arg constructor required by Connect framework.
     assignment = new HashSet<>();
     topicPartitionWriters = new HashMap<>();
+    time = new SystemTime();
```
ewencp: This should be `Time.SYSTEM` so we just share one static instance

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -120,7 +127,7 @@ public void open(Collection<TopicPartition> partitions) {
     assignment.addAll(partitions);
     for (TopicPartition tp : assignment) {
       TopicPartitionWriter writer =
-          new TopicPartitionWriter(tp, storage, writerProvider, partitioner, connectorConfig, context);
+          new TopicPartitionWriter(tp, writerProvider, partitioner, connectorConfig, context, time);
```
ewencp: There's a comment above this `TopicPartitionWriter` constructor that says `visible for testing`. That's no longer true and should be deleted.

There's also some cleanup that seems possible since now the other `TopicPartitionWriter` constructor is only used for testing, has an unused parameter, and when removed could just switch all the tests to using the one used here (they'd just have to pass in `Time.SYSTEM` as one final parameter). But we can always clean that up later.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
```diff
@@ -246,9 +254,152 @@ public void testPreCommit() throws Exception {
...
+
+    verifyOffsets(offsetsToCommit, validOffsets1, context.assignment());
+
+    // 2 hours
+    time.sleep(TimeUnit.HOURS.toMillis(2));
+
+    long[] validOffsets2 = {3, -1};
+
+    // Rotation is only based on rotate.interval.ms, so I need at least one record to trigger flush.
+    task.put(sinkRecords.subList(0, 1));
```
ewencp: This looks a bit weird in the test since you're reusing a record with an old offset. This happens to work because there's no checks on the offsets, but it's actually exercising a scenario that should never happen. You could also just allocate more than you need and use `subList` on each call to `task.put`

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
```diff
@@ -246,9 +254,152 @@ public void testPreCommit() throws Exception {
...
+    long[] validOffsets1 = {-1, -1};
+
+    verifyOffsets(offsetsToCommit, validOffsets1, context.assignment());
+
+    // 1 hour + 10 minutes
+    time.sleep(TimeUnit.HOURS.toMillis(1) + TimeUnit.MINUTES.toMillis(10));
+
+    long[] validOffsets2 = {3, -1};
+
+    // Rotation is only based on rotate.interval.ms, so I need at least one record to trigger flush.
```
ewencp: Comment is leftover from copying old test, but not true here

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
kkonstantine: I can't use the Mock from `org.apache`. That's why I've been forced to use the `io.confluent` in the code and the tests.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
kkonstantine: I was thinking that a PR to clean some unused code would make sense to be issued separately. There are some unused arguments here and there. Since this targets 3.2.x I wanted to keep noise to the very minimum

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
kkonstantine: Good point.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
ewencp: Oh, I guess I never even noticed this. Is there some place we assume the `io.confluent` version? I thought we had kept them pretty close to each other, with the Kafka one more recently getting a minor upgrade re: hi res clocks.

kkonstantine: I could issue a PR on common to synchronize them. We were using `org.apache` in the code and `io.confluent` in the Mock in the tests in this repo, until just the previous PR where I needed to pass the "clock" to the `TopicPartitionWriter` constructor.

kkonstantine:
Thanks for your reviews! Merging.

PR #73, opened 2017-07-20
HOTFIX: Fix flakiness of timebased test.

kkonstantine:


kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -513,11 +513,11 @@ public void testWriteRecordTimeBasedPartitionWallclockMockedWithScheduleRotation
     }
 
     // More records later
-    long timestampLater = time.milliseconds();
     topicPartitionWriter.write();
 
     // 11 minutes later, another scheduled rotation
-    time.sleep(11 * 60 * 1000);
+    time.sleep(TimeUnit.MINUTES.toMillis(11));
+    long timestampLater = time.milliseconds();
```
ewencp: This is the key difference in the test? That we get the `timestampLater` correctly after the `sleep`? Is this related to the fact that we seem to be combining `io.confluent.common.utils.Time` and `org.apache.kafka.common.utils.Time`?

If I'm understanding the bug here, I think this LGTM. But if it's not just the reordering, I might need more explanation...

kkonstantine: Indeed. Looking at the logs the test failed because the path (which encodes the timestamp) was different and not because of the number of records or the offset. One would think the offset doesn't matter in this test, but it matters for the file/key name. (Also I'm glad it's hours and not months because we'd see it failing every other ... NYE :) )

PR #74, opened 2017-07-24
HOTFIX: Fix flaky test by assigning timestamps on write.

kkonstantine:


kkonstantine:
This PR aims to amend the previous fix of this flaky test. The timestamp for records and partitions is determined during a write (meaning during buffering) instead of during the commit of the file. Thus the expected files should have encoded partitions that correspond to the time of writing vs the time of committing.

PR #78, opened 2017-07-31
Retry s3 part upload

niksajakovljevic:
Sometimes due to network glitches S3 part upload fails which results in losing all the previous upload progress  and Kafka Connect Worker Tasks dying. Added support to retry on exceptions caused by these network glitches .


kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -66,6 +65,9 @@
   public static final String AVRO_CODEC_CONFIG = "avro.codec";
   public static final String AVRO_CODEC_DEFAULT = "null";
 
+  public static final String S3_RETRY_CONFIG = "s3.part.retry";
```
ewencp: I wouldn't even add this option, I think retrying should just be the default. If I was going to add an option at all, it'd be the number of retries.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -175,6 +186,60 @@ private MultipartUpload newMultipartUpload() throws IOException {
...
+      } catch(SdkClientException e){
+          cause = e;
+          log.error(errorMsg + ", attempt: "+retries, cause);
+          retries++;
+          try {
+            Thread.sleep(100*2^count);
+          } catch (InterruptedException e1) {
+            log.error("Interrupted while sleeping due to retry", e1);
+          }
+          execute(runnable);
```
ewencp: This should really just be a loop, not a recursive call as using up the stack is unnecessary for this.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -175,6 +186,60 @@ private MultipartUpload newMultipartUpload() throws IOException {
...
+        throw new ConnectException(String.format("Giving up after failing %d times", count), cause);
+      }
+      try {
+        runnable.run();
+      } catch(SdkClientException e){
+          cause = e;
+          log.error(errorMsg + ", attempt: "+retries, cause);
+          retries++;
+          try {
+            Thread.sleep(100*2^count);
```
ewencp: You want the process to bail after the attempt but before sleeping if you've run out of retries. Otherwise you're just unnecessarily sleeping for extra time when you're not going to retry again.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -108,9 +111,17 @@ private void uploadPart(int size) throws IOException {
...
     try {
-      multiPartUpload.uploadPart(new ByteArrayInputStream(buffer.array()), size);
+      if(retry) {
+        RetryWithExponentialBackoff.blocking(new Runnable() {
+          public void run() {
+            multiPartUpload.uploadPart(new ByteArrayInputStream(buffer.array()), partSize);
+          }
+        }, 5, "Part upload failed");
+      } else {
+        multiPartUpload.uploadPart(new ByteArrayInputStream(buffer.array()), partSize);
```
ewencp: I think you can structure this such that it always uses the `RetryWithExponentialBackoff`, no retries would just be the same as `count=1` with the current code.

ewencp:
ok to test

niksajakovljevic:
@ewencp Thanks for the comments. I made some changes accordingly.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -108,9 +111,13 @@ private void uploadPart(int size) throws IOException {
       log.debug("New multi-part upload for bucket '{}' key '{}'", bucket, key);
       multiPartUpload = newMultipartUpload();
     }
-
+    final int partSize = size;
```
niksajakovljevic: Yep, now I see that `size` param is `final`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -56,6 +57,7 @@
   private boolean closed;
   private ByteBuffer buffer;
   private MultipartUpload multiPartUpload;
+  private int retryAttempts;
```
niksajakovljevic: Fixed

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -164,7 +171,7 @@ private ObjectMetadata newObjectMetadata() {
     return meta;
   }
 
-  private MultipartUpload newMultipartUpload() throws IOException {
+  protected MultipartUpload newMultipartUpload() throws IOException {
```
niksajakovljevic: In first place I had an idea of making it accessible for testing, then forgot to change back.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -66,6 +65,9 @@
   public static final String AVRO_CODEC_CONFIG = "avro.codec";
   public static final String AVRO_CODEC_DEFAULT = "null";
 
+  public static final String S3_RETRY_ATTEMPTS_CONFIG = "s3.retry.attempts";
```
niksajakovljevic: Will name it `S3_PART_RETRIES_CONFIG` and default accordingly.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -108,9 +111,13 @@ private void uploadPart(int size) throws IOException {
       log.debug("New multi-part upload for bucket '{}' key '{}'", bucket, key);
       multiPartUpload = newMultipartUpload();
     }
-
+    final int partSize = size;
     try {
-      multiPartUpload.uploadPart(new ByteArrayInputStream(buffer.array()), size);
+      RetryWithExponentialBackoff.blocking(new Runnable() {
```
niksajakovljevic: Yeah, I agree that we can skip the class and have a static method in `S3OutputStream`. In the first run I was thinking to add more to the class (eg. different retry strategies but then switch to some other work related stuff :) )

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -175,6 +182,64 @@ private MultipartUpload newMultipartUpload() throws IOException {
     }
   }
 
+  protected static class RetryWithExponentialBackoff {
```
niksajakovljevic: I commented above for the reasons on going for class implementation, but agree that in the current state of the code it makes sense to have it as a method. On the flip side handling retries is cross cutting concern which was also motivation to have it in a separate class and test it without unnecessary dependencies. Still believe that given number of retry unit tests covers quite a few use cases, but apparently I missed the one with sleep time and thus didn't catch XOR typo.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -175,6 +182,64 @@ private MultipartUpload newMultipartUpload() throws IOException {
...
+      RetryWithExponentialBackoff retry = new RetryWithExponentialBackoff(retries, errorMsg);
+      retry.execute(runnable);
+      return retry;
+    }
+
+    private RetryWithExponentialBackoff execute(Runnable runnable) {
+      do {
+        if (failCount > 0) {
+          try {
+            Thread.sleep(100 * 2 ^ failCount);
```
niksajakovljevic: Good catch! Was doing some math in R lang and continued the same way in Java :) I replaced it with bit shifting.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkTaskTest.java
```diff
@@ -147,6 +153,42 @@ public void testWriteRecordsSpanningMultipleParts() throws Exception {
...
+    localProps.put(S3SinkConnectorConfig.S3_RETRY_ATTEMPTS_CONFIG, "3");
+    setUp();
+
+    // From time to time fail S3 upload part method
+    final AtomicInteger count = new AtomicInteger();
+    PowerMockito.doAnswer(new Answer<UploadPartResult>() {
+      @Override
+      public UploadPartResult answer(InvocationOnMock invocationOnMock) throws Throwable {
+        // Let's fail sometimes. Magic number 20 is picked based on part size and number of records
+        if(count.getAndIncrement() % 20 == 0){
```
niksajakovljevic: Agree. I implemented a formula so it's not empirical based anymore

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -20,6 +20,11 @@
 import com.amazonaws.auth.DefaultAWSCredentialsProviderChain;
 import com.amazonaws.regions.RegionUtils;
 import com.amazonaws.regions.Regions;
+import io.confluent.connect.storage.StorageSinkConnectorConfig;
```
niksajakovljevic: Sorry for this. Will adjust my IDE for future contributions.

niksajakovljevic:
@kkonstantine Thanks for the review comments. I implemented changes accordingly. Please have a look.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -175,6 +182,64 @@ private MultipartUpload newMultipartUpload() throws IOException {
     }
   }
 
+  protected static class RetryWithExponentialBackoff {
+    private int failCount;
+    private int maxRetryAttempts;
```
niksajakovljevic: I believe we need one to keep track of current position and another as a boundary. Having only one counter and decrementing to zero would make it harder(or impossible?) to implement exponential back off. However as I removed the class so these are not class fields any longer.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
kkonstantine: Yes, it's not ideal that you have to diverge from the default ordering of imports. But it makes sense to have it that way. 

To give you a hint, here's what the rules are regarding imports in `codeStyleSettings.xml` when using IntelliJ: 

```
        <option name="IMPORT_LAYOUT_TABLE">
          <value>
            <package name="" withSubpackages="true" static="false" />
            <emptyLine />
            <package name="javax" withSubpackages="true" static="false" />
            <package name="java" withSubpackages="true" static="false" />
            <emptyLine />
            <package name="io.confluent" withSubpackages="true" static="false" />
            <emptyLine />
            <package name="" withSubpackages="true" static="true" />
          </value>
        </option>
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
kkonstantine: So we agree it needs to be reverted? Thanks

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
```diff
@@ -18,6 +18,18 @@
 
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.model.S3ObjectSummary;
+import io.confluent.common.utils.MockTime;
```
kkonstantine: See above the discussion about the ordering of import statements.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -66,6 +65,9 @@
   public static final String AVRO_CODEC_CONFIG = "avro.codec";
   public static final String AVRO_CODEC_DEFAULT = "null";
 
+  public static final String S3_PART_RETRY_CONFIG = "s3.part.retries";
```
niksajakovljevic: No worries, I renamed it

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -153,6 +155,16 @@
                         ++orderInGroup,
                         Width.LONG,
                         "Avro compression codec");
+
+      CONFIG_DEF.define(S3_PART_RETRY_CONFIG,
+                        Type.INT,
+              S3_PART_RETRY_DEFAULT,
```
niksajakovljevic: Sure thing

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -103,14 +106,17 @@ private void uploadPart() throws IOException {
     buffer.clear();
   }
 
-  private void uploadPart(int size) throws IOException {
+  private void uploadPart(final int size) throws IOException {
```
niksajakovljevic: Due to the fact that this argument is later on passed into the inner class (Runnable) it has to stay final.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
niksajakovljevic: Sure thing, guess somehow I missed it ;) fixed now

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
niksajakovljevic: Yes, I addressed some of the shufflings

niksajakovljevic:
@kkonstantine @ewencp  Any plans of merging this ? Would be great to see it in the next release so we can migrate to the new version

kkonstantine:
Thanks @niksajakovljevic. We are definitely on track for the next release re: this PR. 

My last issue was related to the definition of `Runnable` which I find not necessary here. However, I'd be inclined to merge this PR as is and reiterate in a subsequent PR to optimize further. Retries are definitely essential and your contribution is highly appreciated.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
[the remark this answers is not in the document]
kkonstantine: Right, I missed that.

kkonstantine:
Would you mind rebasing on top of master before we are ready to merge? Thanks!

niksajakovljevic:
@kkonstantine No problem, I rebased on top of master now

kkonstantine:
Cool, thanks for your contribution! Merging.

PR #82, opened 2017-08-07
CC-872: Update default part.size number in S3 docs and more.

kkonstantine:
* Bring configuration option up to date for 3.2.2.

### DIFF
--- a/kafka-connect-s3/docs/configuration_options.rst
+++ b/kafka-connect-s3/docs/configuration_options.rst
@@ -133,13 +133,7 @@ Partitioner
   * Type: class
   * Default: io.confluent.connect.storage.partitioner.DefaultPartitioner
   * Importance: high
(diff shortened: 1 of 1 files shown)


kafka-connect-s3/docs/configuration_options.rst
```diff
@@ -135,6 +135,12 @@ Partitioner
   * Importance: high
   * Dependents: ``partition.field.name``, ``partition.duration.ms``, ``path.format``, ``locale``, ``timezone``, ``schema.generator.class``
 
+``schema.generator.class``
```
ewencp: Shall we hide this since we're removing it anyway? It is technically accurate in this version, but not useful.

kkonstantine: Totally. I agree.

Issue #22, opened 2017-03-17
Add support to sync data from Kafka to S3 as it is. 


westerly:
Hi guys,

From what I understand it is not possible to sync schemaless data from Kafka to s3 right now. Would it be possible to support syncing data as it is from Kafka to s3 if the Hive integration is not activated? 

We have an use case in my company where we want to sync plain Strings to S3 so this feature would be awesome to have. Also it would be cool to be able to compress the data synced with Snappy or gzip.

kkonstantine:
Exporting JSON records without schema is already supported with the S3 connector. Exporting binary records "as is" should also be straightforward now that the BinaryArrayConverter has become part of the public set of converters in Kafka (related PR: https://github.com/apache/kafka/pull/2599 and KIP: [128](https://cwiki.apache.org/confluence/display/KAFKA/KIP-128%3A+Add+ByteArrayConverter+for+Kafka+Connect)). 

Once the time-based partitioners are merged, we'll add more specific examples in the quickstart guide to highlight how exporting without schemas works.

jklukas:
> We have an use case in my company where we want to sync plain Strings to S3 so this feature would be awesome to have. Also it would be cool to be able to compress the data synced with Snappy or gzip.

I second this. My team currently maintains our own service for batching up Kafka topics full of JSON-formatted strings and archiving them as newline-delimited gzipped objects in S3 (which Redshift is then able to ingest directly).

I'd love to see an configuration option to compress objects before putting to S3.

jklukas:
> Once the time-based partitioners are merged, we'll add more specific examples in the quickstart guide to highlight how exporting without schemas works.

This would be a great addition.

jklukas:
Time-based partitioners are merged now. I'd still love to see some documentation about exporting without schemas.

jklukas:
See potential implementations of compression and binary data in #83 and #85

jklukas:
ByteArrayFormat support was merged in #85

kkonstantine:
The feature has been added as noted above. Closing

doriwal:
Hi

I am trying to use gzip but its not working , I am reading json events from kafka
confluent version is 4.0.0

{
"connector.class": "io.confluent.connect.s3.S3SinkConnector",
"s3.region": "us-east-1",
"partition.duration.ms": "10",
"flush.size": "3000",
"schema.compatibility": "NONE",
"timezone": "UTC",
"tasks.max": "6",
[...]

PR #83, opened 2017-08-23
[WIP] Add "s3.compression.type" option

jklukas:
As discussed in https://github.com/confluentinc/kafka-connect-storage-cloud/issues/22, there is benefit to being able to compress data before it's shipped to S3.

This PR adds an `s3.compression.type` configuration option that makes use of the [`CompressionType`](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/record/CompressionType.java) class used to configure compression behavior for the producer client. 

At this point, I'm primarily looking for feedback about whether this is a reasonable approach and worth continuing to pursue. I expect there will be some amount of discussion necessary before settling on a solution that will be maintainable and broadly useful.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
@@ -26,12 +26,15 @@ import com.amazonaws.services.s3.model.InitiateMultipartUploadRequest;
 import com.amazonaws.services.s3.model.ObjectMetadata;
 import com.amazonaws.services.s3.model.PartETag;
 import com.amazonaws.services.s3.model.UploadPartRequest;
+import org.apache.kafka.common.record.BufferSupplier;
+import org.apache.kafka.common.record.CompressionType;
+import org.apache.kafka.common.record.RecordBatch;
 import org.apache.kafka.connect.errors.DataException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


jklukas:
[clabot:check]

jklukas:
cc @kkonstantine

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -110,7 +116,9 @@ private void uploadPart(int size) throws IOException {
     }
 
     try {
-      multiPartUpload.uploadPart(new ByteArrayInputStream(buffer.array()), size);
+      InputStream inputStream = compressionType.wrapForInput(
+          buffer, RecordBatch.CURRENT_MAGIC_VALUE, BufferSupplier.NO_CACHING);
```
jklukas: The second and third parameters here look to be ignored for most CompressionType implementations.

I'm mostly focused on the use case of gzip, which I know ignores these two. The gzip format also is tolerant of concatenating several individually zipped streams to a single file. I am not certain that is the case for the other compression types. If that's not true, this approach of zipping each part of the multipart upload separately won't work. If it makes more sense to scope this to gzip and not support the other formats, I'm open to making that change.

kkonstantine:
Thanks @jklukas for the initiative. 
Marking this PR as Work-In-Progress given its state and the type of feedback you requested. 

Seems pretty useful for several people, will try to give my feedback soon!

jklukas:
> Marking this PR as Work-In-Progress given its state and the type of feedback you requested.

Sounds reasonable.

jklukas:
I've realized that my approach for compression here is inverted. We need to `wrapForOutput` rather than input in order to write compressed data. Will be looking at refactoring.

jklukas:
Will open a completely separate PR once I have a better idea of how to approach this.

imduffy15:
What is the status of this?

jklukas:
See https://github.com/confluentinc/kafka-connect-storage-cloud/pull/98

PR #84, opened 2017-08-23
Create CONTRIBUTING.md

jklukas:
It would be nice to avoid the back-and-forth of having clabot comment on a new PR in order for folks to know about the CLA. A CONTRIBUTING.md file gives some heads-up about that step so that folks can take care of it before submitting their first PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,7 @@
+In order for us to consider merging a contribution, you will need to sign our
+**C**ontributor **L**icense **A**greement.
+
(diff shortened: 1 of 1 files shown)


jklukas:
> But I wouldn't remove the bot. People (like myself) are often too impatient to read carefully all the docs :)

I certainly agree with keeping the bot in place. I'm the kind of person who loves knowing what I'm getting into before hitting the "Submit" button, so this should let us have it both ways.

kkonstantine:
Sounds good! I'm merging this one, since it's a good starting point for a `CONTRIBUTING.md` doc.

PR #85, opened 2017-08-24
Add ByteArrayFormat

jklukas:
As @kkonstantine suggested in https://github.com/confluentinc/kafka-connect-storage-cloud/issues/22#issuecomment-287885581, availability of `ByteArrayConverter` since 0.11.0.0 allows us to skip parsing of message values completely and write them as is.

This PR adds a `ByteArrayFormat` and `ByteArrayRecordWriterProvider` to wrap the new converter.

I've marked this WIP as I'd like to gather some feedback about whether there are other potential approaches and whether there are additional options we'd need to see for this to be mergeable.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayFormat.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayFormat.java
@@ -50,7 +50,8 @@ public class ByteArrayFormat implements Format<S3SinkConnectorConfig, String> {
 
   @Override
   public HiveFactory getHiveFactory() {
-    throw new UnsupportedOperationException("Hive integration is not currently supported in S3 Connector");
+    throw new UnsupportedOperationException(
+        "Hive integration is not currently supported in S3 Connector");
   }
 
 }
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayFormat.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayFormat.java
@@ -35,9 +35,6 @@ public class ByteArrayFormat implements Format<S3SinkConnectorConfig, String> {
     this.storage = storage;
     this.converter = new ByteArrayConverter();
     Map<String, Object> converterConfig = new HashMap<>();
-    converterConfig.put("schemas.enable", "false");
-    converterConfig.put("schemas.cache.size",
-        String.valueOf(storage.conf().get(S3SinkConnectorConfig.SCHEMA_CACHE_SIZE_CONFIG)));
     this.converter.configure(converterConfig, false);
   }
 
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayFormat.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayFormat.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2017 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.connect.s3.format.bytearray;
+
+import io.confluent.connect.s3.S3SinkConnectorConfig;
+import io.confluent.connect.s3.storage.S3Storage;
+import io.confluent.connect.storage.format.Format;
+import io.confluent.connect.storage.format.RecordWriterProvider;
+import io.confluent.connect.storage.format.SchemaFileReader;
+import io.confluent.connect.storage.hive.HiveFactory;
+import org.apache.kafka.connect.converters.ByteArrayConverter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ByteArrayFormat implements Format<S3SinkConnectorConfig, String> {
+  private final S3Storage storage;
+  private final ByteArrayConverter converter;
+
+  public ByteArrayFormat(S3Storage storage) {
+    this.storage = storage;
+    this.converter = new ByteArrayConverter();
+    Map<String, Object> converterConfig = new HashMap<>();
+    converterConfig.put("schemas.enable", "false");
+    converterConfig.put("schemas.cache.size",
+        String.valueOf(storage.conf().get(S3SinkConnectorConfig.SCHEMA_CACHE_SIZE_CONFIG)));
+    this.converter.configure(converterConfig, false);
+  }
+
+  @Override
+  public RecordWriterProvider<S3SinkConnectorConfig> getRecordWriterProvider() {
+    return new ByteArrayRecordWriterProvider(storage, converter);
+  }
+
+  @Override
+  public SchemaFileReader<S3SinkConnectorConfig, String> getSchemaFileReader() {
+    throw new UnsupportedOperationException("Reading schemas from S3 is not currently supported");
+  }
+
+  @Override
+  public HiveFactory getHiveFactory() {
+    throw new UnsupportedOperationException("Hive integration is not currently supported in S3 Connector");
+  }
+
+}
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
@@ -29,23 +29,26 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.Arrays;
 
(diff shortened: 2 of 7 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+public class ByteArrayRecordWriterProvider implements RecordWriterProvider<S3SinkConnectorConfig> {
+
+  private static final Logger log = LoggerFactory.getLogger(ByteArrayRecordWriterProvider.class);
+  private static final String EXTENSION = ".out";
```
jklukas: I don't know the right approach for extension. `.out` is intended as a placeholder for now.

My use case for this format is that I have a topic with schema-less JSON messages and I want to write them unchanged and unparsed to S3. So Kafka Connect can treat the messages as byte arrays and never know they're JSON, but I'd like the output file to have a ".json" extension.

Has there been any thought about making the extension configurable for formats?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
```diff
@@ -0,0 +1,83 @@
...
+  private static final String LINE_SEPARATOR = System.lineSeparator();
```
jklukas: System line separator works for my use case, but others may have byte array use cases that could contain a literal newline and they might want to use a different separator.

Should line separator be configurable? Should it also be configurable for JsonFormat?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
```diff
@@ -0,0 +1,83 @@
...
+  public RecordWriter getRecordWriter(final S3SinkConnectorConfig conf, final String filename) {
+    return new RecordWriter() {
+      final S3OutputStream s3out = storage.create(filename, true);
+
+      @Override
+      public void write(SinkRecord record) {
+        log.trace("Sink record: {}", record);
+        try {
+          byte[] bytes = converter.fromConnectData(
+              record.topic(), record.valueSchema(), record.value());
```
jklukas: My thought here is to pass the `record.value()` blindly into the converter without verifying that it's a byte array since `fromConnectData` contains checks around type and will emit a DataException we catch below.

jklukas:
@klamouri - Do you have any opinions on the implementation here?

klamouri:
Well I'm not affiliated to Confluent so my opinion has little weight but looking at your code I feel like there is nothing aberrant. However I think you're right making the extension and the separator configurable would be the best ;)

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayFormat.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class ByteArrayFormat implements Format<S3SinkConnectorConfig, String> {
+  private final S3Storage storage;
+  private final ByteArrayConverter converter;
+
+  public ByteArrayFormat(S3Storage storage) {
+    this.storage = storage;
+    this.converter = new ByteArrayConverter();
+    Map<String, Object> converterConfig = new HashMap<>();
+    converterConfig.put("schemas.enable", "false");
```
kkonstantine: Not required by this converter.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayFormat.java
```diff
@@ -0,0 +1,59 @@
...
+    converterConfig.put("schemas.cache.size",
```
kkonstantine: Not required by this converter.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
kkonstantine: I'd use `.bin` as a default. 

And also, I think we enabled the way to define a "user-defined" per formatter config property. Although we want to be prudent and not add too many configuration parameters, maybe it would make sense to add this one only for this formatter (not in general for any formatter).

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
kkonstantine: Same approach as with the extension. Formatter specific configuration.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
```diff
@@ -0,0 +1,83 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+public class ByteArrayRecordWriterProvider implements RecordWriterProvider<S3SinkConnectorConfig> {
+
+  private static final Logger log = LoggerFactory.getLogger(ByteArrayRecordWriterProvider.class);
+  private static final String EXTENSION = ".out";
+  private static final String LINE_SEPARATOR = System.lineSeparator();
+  private static final byte[] LINE_SEPARATOR_BYTES = LINE_SEPARATOR.getBytes();
```
kkonstantine: You'd only need one member field variable here: 
```
private static final byte[] LINE_SEPARATOR = System.lineSeparator().getBytes();
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
jklukas: > I think we enabled the way to define a "user-defined" per formatter config property

Can you give me any more context on this? I'm not sure if you're referring to an upcoming change or something that already exists.

jklukas:
Thanks for the feedback, @kkonstantine.

> We'd need tests for sure.

Definitely. There's a good bit of indirection in the tests, so I'm still getting a handle on how much needs to add tests for this format. If you have any guidance there about what you expect to see in the tests, that would be helpful, but I will indeed take a swing at adding some tests.

jklukas:
I'll remove the WIP label and ask for further review once I've gotten the tests added and made the changes @kkonstantine suggested.

kkonstantine:
I guess a good way to start would be to see what are the tests for json format, and write the equivalent cases if it makes sense. Also think if there's something specific to this format (especially if we go with adding its special configs). Thanks!

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
jklukas: It looks like it's going to be a bit tortured to apply config values to `ByteArrayRecordWriterProvider` since the `conf` doesn't get passed in to the constructor. I could add a param to the constructor to pass in the conf, but that seems to be getting a bit redundant.

Perhaps again this is due to missing context about how user-defined per formatter configs are imagined to work.

kkonstantine:
Lots of progress. Thanks! Is this still WIP? Or is it ready for another complete review?

kkonstantine:
ok to test

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterByteArrayTest.java
```diff
@@ -0,0 +1,228 @@
...
+                            "{\"type\":\"boolean\",\"optional\":true,\"field\":\"booleanField\"}," +
+                            "{\"type\":\"int32\",\"optional\":true,\"field\":\"intField\"}," +
+                            "{\"type\":\"int64\",\"optional\":true,\"field\":\"longField\"}," +
+                            "{\"type\":\"string\",\"optional\":false,\"field\":\"stringField\"}]," +
+                            "\"payload\":" +
+                            "{\"booleanField\":\"true\"," +
+                            "\"intField\":" + String.valueOf(ibase) + "," +
+                            "\"longField\":" + String.valueOf((long) ibase) + "," +
+                            "\"stringField\":str" + String.valueOf(ibase) +
+                            "}}").getBytes();
```
jklukas: I copied this test verbatim from the JSON tests, so this message doesn't make much sense here.

I tried out a change where instead of this message, we just fill a byte array with random bytes (and make sure not to include the separator bytes). I think that exposed a logic error in the existing JSON tests. This is structured with TopicPartition as the inner loop, but `verify` puts TopicPartition as the outer loop, so expectedRecords and the reconstituted records don't match up.

It looks like the JSON tests only succeed because all the messages have the same value.

Do you have any opinions about what payloads we should be generating here to fully exercise the new format? Is the test as written right now fine for the scope of this PR? Or would you prefer that we go ahead and modify the JSON test logic inside this PR and factor out this method to be shared between the JSON and byte array tests?

jklukas:
Removed the WIP label and left a question about tests.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterByteArrayTest.java
kkonstantine: From what I can remember having the same message was an intentional tradeoff. We should change this in both cases, but that happen in a follow up PR.

jklukas:
Thanks again for the thoughtful review, @kkonstantine and moving this through to getting to a mergeable state.

PR #87, opened 2017-09-07
Add s3.canned.acl config

jklukas:
This option is helpful for writing cross account.

In our use case, our Kafka Connect instance runs in a separate account from where our destination S3 buckets live. We want to be able to specify `BucketOwnerFullControl` so that the destination account has read and write access to the objects Kafka Connect is writing. The default `Private` acl allows access only from the account that originally wrote the data.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -20,6 +20,7 @@ import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.auth.DefaultAWSCredentialsProviderChain;
 import com.amazonaws.regions.RegionUtils;
 import com.amazonaws.regions.Regions;
+import com.amazonaws.services.s3.model.CannedAccessControlList;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
@@ -29,6 +30,7 @@ import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.connect.errors.ConnectException;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -62,6 +64,9 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
 
   public static final String REGION_CONFIG = "s3.region";
   public static final String REGION_DEFAULT = Regions.DEFAULT_REGION.getName();
+
+  public static final String ACL_CANNED_CONFIG = "s3.acl.canned";
+  public static final String ACL_CANNED_DEFAULT = null;
 
   public static final String AVRO_CODEC_CONFIG = "avro.codec";
   public static final String AVRO_CODEC_DEFAULT = "null";
@@ -136,6 +141,17 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
                         Width.LONG,
                         "S3 Server Side Encryption Algorithm");
 
+      CONFIG_DEF.define(ACL_CANNED_CONFIG,
+                        Type.STRING,
+                        ACL_CANNED_DEFAULT,
+                        new CannedAclValidator(),
+                        Importance.LOW,
+                        "An S3 canned ACL header value to apply when writing objects.",
+                        group,
+                        ++orderInGroup,
+                        Width.LONG,
+                        "S3 Canned ACL");
+
       CONFIG_DEF.define(WAN_MODE_CONFIG,
                         Type.BOOLEAN,
                         WAN_MODE_DEFAULT,
@@ -203,6 +219,10 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
 
   public String getSsea() {
     return getString(SSEA_CONFIG);
+  }
+
+  public CannedAccessControlList getCannedAcl() {
+    return CannedAclValidator.ACLS_BY_HEADER_VALUE.get(getString(ACL_CANNED_CONFIG));
   }
 
   public int getPartSize() {
@@ -317,6 +337,36 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
     @Override
     public String toString() {
       return "[" + Utils.join(RegionUtils.getRegions(), ", ") + "]";
+    }
+  }
+
+  private static class CannedAclValidator implements ConfigDef.Validator {
+    public static final Map<String, CannedAccessControlList> ACLS_BY_HEADER_VALUE = new HashMap<>();
+    public static final String ALLOWED_VALUES;
+
+    static {
+      List<String> aclHeaderValues = new ArrayList<>();
+      for (CannedAccessControlList acl : CannedAccessControlList.values()) {
+        ACLS_BY_HEADER_VALUE.put(acl.toString(), acl);
+        aclHeaderValues.add(acl.toString());
+      }
+      ALLOWED_VALUES = Utils.join(aclHeaderValues, ", ");
+    }
+
+    @Override
+    public void ensureValid(String name, Object cannedAcl) {
+      if (cannedAcl == null) {
+        return;
+      }
+      String aclStr = ((String) cannedAcl).trim();
+      if (!ACLS_BY_HEADER_VALUE.containsKey(aclStr)) {
+        throw new ConfigException(name, cannedAcl, "Value must be one of: " + ALLOWED_VALUES);
+      }
+    }
+
+    @Override
+    public String toString() {
+      return "[" + ALLOWED_VALUES + "]";
     }
   }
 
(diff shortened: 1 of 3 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -63,6 +65,9 @@
   public static final String REGION_CONFIG = "s3.region";
   public static final String REGION_DEFAULT = Regions.DEFAULT_REGION.getName();
 
+  public static final String CANNED_ACL_CONFIG = "s3.canned.acl";
+  public static final String CANNED_ACL_DEFAULT = "Private";
```
kkonstantine: We should compare against the `toString` representation of the enum (`cannedAclHeader` internally) not the enum name. So, `private` and `bucket-owner-read` instead of `Private` and `BucketOwnerRead`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -320,6 +341,33 @@ public String toString() {
     }
   }
 
+  private static class CannedAclValidator implements ConfigDef.Validator {
+    private static String allowedValues;
+
+    static {
+      List<String> aclStrings = new ArrayList<>();
+      for (CannedAccessControlList acl: CannedAccessControlList.values()) {
+        aclStrings.add(acl.name());
```
kkonstantine: `toString` instead

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -136,6 +141,18 @@
                         Width.LONG,
                         "S3 Server Side Encryption Algorithm");
 
+      CONFIG_DEF.define(CANNED_ACL_CONFIG,
+                        Type.STRING,
+                        CANNED_ACL_DEFAULT,
+                        new CannedAclValidator(),
+                        Importance.LOW,
+                        "The S3 canned ACL to apply when writing objects. " +
+                            "by default, 'CannedAccessControlList.Private'",
```
kkonstantine: `+` should be on the following line, or else checkstyle will complain.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -320,6 +341,33 @@ public String toString() {
...
+        aclStrings.add(acl.name());
+      }
+      allowedValues = Utils.join(aclStrings, ", ");
+    }
+
+    @Override
+    public void ensureValid(String name, Object cannedAcl) {
+      String aclStr = ((String) cannedAcl).trim();
+      try {
+        CannedAccessControlList.valueOf(aclStr);
```
kkonstantine: we need to compare against the actual values not the enum names. We'll need to reuse the `Map<String, CannedAccessControlList>` that we'll need for `getCannedAcl ` or a `Set<String>` to store the allowed values.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -205,6 +222,10 @@ public String getSsea() {
     return getString(SSEA_CONFIG);
   }
 
+  public String getCannedAcl() {
+    return getString(CANNED_ACL_CONFIG);
```
kkonstantine: we need a mapping from string to enum.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -65,6 +67,7 @@ public S3OutputStream(String key, S3SinkConnectorConfig conf, AmazonS3 s3) {
     this.key = key;
     this.ssea = conf.getSsea();
     this.partSize = conf.getPartSize();
+    this.cannedAcl = CannedAccessControlList.valueOf(conf.getCannedAcl());
```
kkonstantine: This won't work. Also better to include the logic within `conf.getCannedAcl()`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
kkonstantine: Also, I'm thinking that for complete backwards compatibility, maybe the default should be `null` here, because that's what it is so far.

jklukas:
These suggestions all sound reasonable. Will fix.

Is there any CI for this project? It would be nice to verify that tests are passing centrally.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -205,6 +222,10 @@ public String getSsea() {
     return getString(SSEA_CONFIG);
   }
 
+  public CannedAccessControlList getCannedAcl() {
+    return CannedAccessControlList.valueOf(CANNED_ACL_CONFIG);
```
jklukas: I need to come back to making a map that gets us from toString values to `name` values.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -63,6 +65,9 @@
   public static final String REGION_CONFIG = "s3.region";
   public static final String REGION_DEFAULT = Regions.DEFAULT_REGION.getName();
 
+  public static final String ACL_CANNED_CONFIG = "s3.acl.canned";
+  public static final String ACL_CANNED_DEFAULT = CannedAccessControlList.Private.toString();
```
jklukas: Handling the default as null adds significant additional complexity. I'm open to continuing down that path if you think the additional complexity is worth it, @kkonstantine.

jklukas:
Comments are now addressed and existing tests are passing.

I'm going to look into adding a test or two to exercise this code, then will ask for more review.

jklukas:
Added a test which exercises a config value other than the default.

https://github.com/findify/s3mock specifically calls out ACLs as a non-implemented feature. Any ideas for more thoroughly testing this, @kkonstantine, or does the test here look sufficient?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -63,6 +65,9 @@
   public static final String REGION_CONFIG = "s3.region";
   public static final String REGION_DEFAULT = Regions.DEFAULT_REGION.getName();
 
+  public static final String ACL_CANNED_CONFIG = "s3.acl.canned";
+  public static final String ACL_CANNED_DEFAULT = null;
```
jklukas: After implementing the map of header values to ACLs, the change to null for default actually became straight-forward, so I committed that change.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -320,6 +340,36 @@ public String toString() {
...
+        aclHeaderValues.add(acl.toString());
+      }
+      allowedValues = Utils.join(aclHeaderValues, ", ");
+    }
+
+    @Override
+    public void ensureValid(String name, Object cannedAcl) {
+      if (cannedAcl == null) {
+        return;
+      }
```
jklukas: Let me know if you'd prefer a style other than an early return for the null case.

jklukas:
@kkonstantine - I'm all done with changes. This is ready for review again.

kkonstantine:
ok to test

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -320,6 +340,36 @@ public String toString() {
     }
   }
 
+  private static class CannedAclValidator implements ConfigDef.Validator {
+    public static Map<String, CannedAccessControlList> aclsByHeaderValue = new HashMap<>();
```
kkonstantine: Cool. Final brush strokes, let's make those two `final` and then comply with the format for `static final` member fields. That would be `ACLS_BY_HEADER_VALUE` and `ALLOWED_VALUES` respectively.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -320,6 +340,36 @@ public String toString() {
     }
   }
 
+  private static class CannedAclValidator implements ConfigDef.Validator {
+    public static final Map<String, CannedAccessControlList> ACLS_BY_HEADER_VALUE = new HashMap<>();
+    public static final String ALLOWED_VALUES;
```
jklukas: These are both now `final` and renamed to all caps. I also made ALLOWED_VALUES public.

jklukas:
Addressed the two final comments and did a rebase/squash on master.

Thanks much for the thoughtful review, @kkonstantine.

kkonstantine:
Testing looks ok for now given that the mock doesn't do anything special for ACLs. I'll go ahead and merge this one. Thanks for the patch @jklukas !

PR #91, opened 2017-09-15
added s3.sse.aws.kms.id to specify a particular key in KMS to use

AlexPiermatteo:
At the moment is not possible to specify a particular master key in KMS to use when we want to encrypt data on a server side into S3 (SSE:KMS).

Through the _s3.ssea.name_ config is only possible to specify _"aws:kms"_ as encryption type, but then only the default key in KMS is used.
I added a new property to handle this and give the possibility to specify a particular master key that already exists in KMS (either using the id or the alias name of the master key).

### DIFF
--- a/kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
+++ b/kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
@@ -220,9 +220,10 @@ public class TopicPartitionWriterTest extends TestWithMockedS3 {
     setUp();
 
     long timeBucketMs = TimeUnit.SECONDS.toMillis(5);
+    long partitionDurationMs = TimeUnit.MINUTES.toMillis(1);
     // Define the partitioner
     Partitioner<FieldSchema> partitioner = new TimeBasedPartitioner<>();
-    parsedConfig.put(PartitionerConfig.PARTITION_DURATION_MS_CONFIG, timeBucketMs);
+    parsedConfig.put(PartitionerConfig.PARTITION_DURATION_MS_CONFIG, partitionDurationMs);
     parsedConfig.put(PartitionerConfig.SCHEMA_GENERATOR_CLASS_CONFIG, TimeBasedSchemaGenerator.class);
     parsedConfig.put(
         PartitionerConfig.PATH_FORMAT_CONFIG, "'year'=YYYY_'month'=MM_'day'=dd_'hour'=HH_'min'=mm_'sec'=ss");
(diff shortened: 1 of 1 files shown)


AlexPiermatteo:
[clabot:check]

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -166,6 +163,10 @@ private ObjectMetadata newObjectMetadata() {
 
   private MultipartUpload newMultipartUpload() throws IOException {
     InitiateMultipartUploadRequest initRequest = new InitiateMultipartUploadRequest(bucket, key, newObjectMetadata());
+
+    if (ssea.equals("aws:kms") && StringUtils.isNotBlank(awsKmsKeyId))
+      initRequest.withSSEAwsKeyManagementParams(new SSEAwsKeyManagementParams(awsKmsKeyId));
```
jklukas: This method returns an updated InitiateMultipartUploadRequest rather than mutating the value, so this line should be assigning the result to `initRequest`.

http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/InitiateMultipartUploadRequest.html#withSSEAwsKeyManagementParams-com.amazonaws.services.s3.model.SSEAwsKeyManagementParams-

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -20,12 +20,7 @@
 import com.amazonaws.event.ProgressEvent;
 import com.amazonaws.event.ProgressListener;
 import com.amazonaws.services.s3.AmazonS3;
-import com.amazonaws.services.s3.model.AbortMultipartUploadRequest;
-import com.amazonaws.services.s3.model.CompleteMultipartUploadRequest;
-import com.amazonaws.services.s3.model.InitiateMultipartUploadRequest;
-import com.amazonaws.services.s3.model.ObjectMetadata;
-import com.amazonaws.services.s3.model.PartETag;
-import com.amazonaws.services.s3.model.UploadPartRequest;
+import com.amazonaws.services.s3.model.*;
```
jklukas: Maintainers will likely want these import changes reverted.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
jklukas: Mutation in place would probably be cleaner here, so replacing this call with `setSSEAwsKeyManagementParams`.

kkonstantine:
I changed the base to master, but you'll need to rebase also on top of master instead of the `CC-500-Provide-exactly-once-time-based-partitioning-in-S3` branch.

PR #96, opened 2017-09-25
Allow S3OutputStream to commit as part of close()

jklukas:
The existing RecordWriterProvider classes in this project need to be
careful to flush their writers before calling commit() on the
S3OutputStream and include comments warning about this need.

This PR moves the commit logic directly into the close() method
so that closing a wrapper can more elegantly initiate the multipart
upload.

I'm particularly interested in this change as it paves the way
for an elegant implementation of compression in this project.
In particular, the current implementation makes it necessary to
call GZIPOutputStream.finish() before closing, which is a method
specific to DeflatorOutputStream. If commit can happen inside close(),
then we can handle compressors more generically via the OutputStream
interface.

For review, I'm looking for feedback about whether this approach actually seems like an improvement or if it could cause problems that I'm not anticipating. I will admit that I don't have deep experience with Java I/O streams and documentation on expected usage is a bit slim. I'm inferring, though, that it's generally expected that wrappers should allow flushing and closing to cascade down to the streams they wrap.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
@@ -40,6 +40,7 @@ import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Output stream enabling multi-part uploads of Kafka records.
@@ -58,6 +59,7 @@ public class S3OutputStream extends OutputStream {
   private ByteBuffer buffer;
   private MultipartUpload multiPartUpload;
   private final int retries;
+  private final AtomicBoolean commitOnCloseEnabled = new AtomicBoolean(false);
 
   public S3OutputStream(String key, S3SinkConnectorConfig conf, AmazonS3 s3) {
     this.s3 = s3;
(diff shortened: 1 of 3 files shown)


jklukas:
cc @kkonstantine for thoughts on whether this looks like a clear improvement.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -131,29 +133,14 @@ public void run() {
...
-    if (closed) {
-      log.warn(
-          "Tried to commit data for bucket '{}' key '{}' on a closed stream. Ignoring.",
-          bucket,
-          key
-      );
-      return;
-    }
-
-    try {
```
jklukas: This try block is moved verbatim into `close()`

kkonstantine:
Closing the stream does not always imply an intentional call on `close` that signifies normal operation. The stream might close on an error as well, and committing the parts and the offsets at that point would be wrong. 

These two operations were decoupled on purpose since the initial implementation of this connector.
Given that, I don't think we should merge this PR.

jklukas:
> These two operations were decoupled on purpose since the initial implementation of this connector

Understood. This PR maintains the idea that `close` by default handles the error case and aborts the upload. An intentional call to `enableCommitOnClose()` is required to cause the upload to finish. I see this as essentially swapping a required call to `commit()` with a required call to `enableCommitOnClose()`.

That said, I respect the decision if you feel the complete separation of `commit()` provides better safety. Feel free to close if you don't see more to discuss here.

jklukas:
Closed this based on https://github.com/confluentinc/kafka-connect-storage-cloud/pull/96#issuecomment-331917759. Went ahead with implementing a gzip compression option using the existing `commit()` interface in https://github.com/confluentinc/kafka-connect-storage-cloud/pull/98

PR #97, opened 2017-09-26
MINOR: S3 Proxy Settings revisited

kkonstantine:


kkonstantine:
Issuing a new PR to continue the work introduced with https://github.com/confluentinc/kafka-connect-storage-cloud/pull/30

The suggestion here is that all the proxy info will be encoded as a url in `s3.proxy.url` with the option to hide the username and password with separate config properties (`s3.proxy.user` and `s3.proxy.password` respectively). If a user want to include the user/pass unhidden in the url they may user the standard encoding of the url and leave the other two properties unset.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ProxyConfig.java
```diff
@@ -0,0 +1,105 @@
...
+    } catch (MalformedURLException e) {
+      throw new ConfigException(
+          "Invalid "
+              + S3_PROXY_URL_CONFIG
+              + " property: '"
+              + config.getString(S3_PROXY_URL_CONFIG)
+              + "'"
+      );
+    }
+  }
```
rhauch: This is non-trivial, so we probably should have a unit test class for this that verifies the behavior with valid and unexpected values.

kkonstantine: The unit test is under way. But due to the temporary time difference I wanted to give an early heads up.

kkonstantine: The unit test is under way. But due to the temporary time difference I wanted to give an early heads up.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -217,6 +227,36 @@
...
+                        S3_PROXY_URL_DEFAULT,
+                        Importance.LOW,
+                        "S3 Proxy settings encoded in URL syntax",
+                        group,
+                        ++orderInGroup,
+                        Width.LONG,
+                        "S3 Proxy Settings");
+
+      CONFIG_DEF.define(S3_PROXY_USER_CONFIG,
+                        Type.PASSWORD,
```
rhauch: Should this be `Type.STRING` for the user?

kkonstantine: I felt that the username is also sensitive information to dump to the logs (same as the password). But that's up for discussion. /cc @ewencp

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -217,6 +227,36 @@
                         ++orderInGroup,
                         Width.LONG,
                         "Line separator ByteArrayFormat");
+
+      CONFIG_DEF.define(S3_PROXY_URL_CONFIG,
+                        Type.STRING,
+                        S3_PROXY_URL_DEFAULT,
+                        Importance.LOW,
+                        "S3 Proxy settings encoded in URL syntax",
```
rhauch: Might be nice to have a bit more description on these 3 new configuration properties to help explain when/why these would be used, and how they differ from regular S3 credentials. Importance is low, so they will at least be separated from the existing credentials.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
rhauch: Usernames that require authentication are not usually sensitive, whereas knowing having the username in the logs helps diagnose authentication/authorization issues.

kkonstantine:
ok to test

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
kkonstantine: Added more details.

Also changed username to be String (can't reply in line to the hidden comment below).

kkonstantine:
@rhauch I believe I've addressed your latest comments. Feel free to take a look this is correct.

kkonstantine:
Thanks for the reviews @rhauch. Merging!

PR #100, opened 2017-09-28
CC-1167: Add recommenders for class type properties

kkonstantine:


kkonstantine:
Requires https://github.com/confluentinc/kafka-connect-storage-common/pull/36

kkonstantine:
ok to test

kkonstantine:
Now that the required changes have been merged to `kafka-connect-storage-common` this PR turned green and is ready for review. Thanks!

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -154,6 +162,25 @@
                         Width.LONG,
                         "Avro compression codec");
     }
+
+    FORMAT_CLASS_RECOMMENDER.addValidValues(
```
ewencp: hmm, it seems like this would result in weird suggestions if the plugins are loaded via classpath instead of `plugin.path` since this static field will be used by both s3 and hdfs connectors? same for the other fields as well.

kkonstantine: Strange. I thought I mentioned this already, but I don't see it in any of the comments in the other PR. 

Yes, that's a consequence of this approach. Once both connectors are added, all the formats will show up and one will have to pay attention to the actual package to select the appropriate class (that's why I used `Set` instead of `List` in the base PR, so if the connectors use exactly the same class, the class won't be added to the recommender twice).

ewencp: Couldn't we just make the `CONFIG_DEF` and these recommenders in `StorageSinkConnectorConfig` non-`static` so they are properly isolated?

kkonstantine: Besides this entailing pretty wide changes in the configs code (here both repos), I think the main barrier is related to validation. 

Most connectors would create their configs during the call to `start` method. By requiring `Connector.config` not to call a static method to get a `ConfigDef` but a non-static one, we'd need to make sure at the framework that every call to `validate` or `config` is preceded by a call to `start`. I'm sensing this is not practical at this point and for the given change. Besides that, there might be other implications that are not obvious to me right now. 

The fact is that the pattern has been to declare static `ConfigDef`s across the board and I feel that a divergence from this pattern is risky at this point (and for the requirements of this change). 

Let me know what you think.

rhauch: > Once both connectors are added, all the formats will show up and one will have to pay attention to the actual package to select the appropriate class

Can we really rely upon the user to do the right thing? This seems quite a subtle distinction and a potential usability problem.

rhauch: > Most connectors would create their configs during the call to start method. By requiring Connector.config not to call a static method to get a ConfigDef but a non-static one ...

What about keeping the method static but instead having the method create a new `ConfigDef` each time the method is called? That shouldn't really change the behavior (or at least it would reduce the impact to cases that are assuming a shared instance), and it would prevent sharing. And while the base class' `ConfigDef` would not be shared, each connector's `ConfigDef` could still remain static. WDYT?

kkonstantine:
@rhauch allow me to return the question and ask you, if we needed to pass 2 recommenders what would you do instead?

The answer to your second question is that because this config has already most of its configs as non-static members, and now that the underlying base config does not have a static `CONFIG_DEF` I want to make this more uniform. In that sense calling `newConfigDef` will make sense for everything and not only some of the configs.

rhauch:
If Athe function needdd two Recommenders, I’d have the signature be `newConfigDef(Recommender formatClassRecommender, Recommender storageClassRecommender)`.

kkonstantine:
With already a few class type properties and a few dozen properties overall, adding every recommender as a separate argument is not an obvious scalable solution. Map would be another option. 

Anyway, we'll cross this bridge when we come to it. Adding as simple argument at the moment.

kkonstantine:
Thanks for the reviews! Merging.

PR #102, opened 2017-10-02
Allow to write records at the root of the bucket

klamouri:
Hi,

It would be useful to allow the s3 connector to write at the root of the bucket. It's fairly straightforward to implement this with the existing configs.

If topic.dir is either set with a blank value or a slash the connector should write at the root of the bucket

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -16,6 +16,7 @@
 
 package io.confluent.connect.s3;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.data.Schema;
@@ -434,7 +435,11 @@ public class TopicPartitionWriter {
   }
 
   private String fileKey(String topicsPrefix, String keyPrefix, String name) {
-    return topicsPrefix + dirDelim + keyPrefix + dirDelim + name;
+    String base = "";
+    if (StringUtils.isNotBlank(topicsPrefix) && !topicsPrefix.equals("/")) {
+      base = topicsPrefix + dirDelim;
+    }
+    return base + keyPrefix + dirDelim + name;
   }
 
   private String fileKeyToCommit(String dirPrefix, long startOffset) {
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -437,8 +437,8 @@ public class TopicPartitionWriter {
   private String fileKey(String topicsPrefix, String keyPrefix, String name) {
     String suffix = keyPrefix + dirDelim + name;
     return StringUtils.isNotBlank(topicsPrefix)
-            ? topicsPrefix + dirDelim + suffix
-            : suffix;
+           ? topicsPrefix + dirDelim + suffix
+           : suffix;
   }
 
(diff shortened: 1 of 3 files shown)


klamouri:
[clabot:check]

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -16,6 +16,7 @@
 
 package io.confluent.connect.s3;
 
+import org.apache.commons.lang3.StringUtils;
```
kkonstantine: Would you mind using `io.confluent.connect.storage.common.util.StringUtils;` instead. 

I know it's tricky because they have the same class name and ... they do the same thing :) But saves us one external dependency (even if commons.lang3 is already brought in).

kafka-connect-s3/src/test/java/io/confluent/connect/s3/util/FileUtils.java
```diff
@@ -16,14 +16,19 @@
 
 package io.confluent.connect.s3.util;
 
+import org.apache.commons.lang3.StringUtils;
```
kkonstantine: Again,  preferably: `import io.confluent.connect.storage.common.util.StringUtils;`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -434,7 +435,11 @@ private String getCommitFilename(String encodedPartition) {
   }
 
   private String fileKey(String topicsPrefix, String keyPrefix, String name) {
-    return topicsPrefix + dirDelim + keyPrefix + dirDelim + name;
+    String base = "";
+    if (StringUtils.isNotBlank(topicsPrefix) && !topicsPrefix.equals("/")) {
```
kkonstantine: Seems we might be overly specific by allowing everything besides `/`. If someone set their `topics.dir` property to `/` they might know what they're doing (In S3 this produces a different key than the empty string obviously, but I don't it should be on us to prevent this here).

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -434,7 +435,11 @@ private String getCommitFilename(String encodedPartition) {
...
+      base = topicsPrefix + dirDelim;
+    }
+    return base + keyPrefix + dirDelim + name;
```
kkonstantine: Also, I apologise for being overly specific here, but this might be one of the rare cases where a sightly lengthy ternary operator might be more readable than a block of 3-4 lines. 

How about?

```java
private String fileKey(String topicsPrefix, String keyPrefix, String name) {
    String suffix = keyPrefix + dirDelim + name;
    return StringUtils.isNotBlank(topicsPrefix)
           ? topicsPrefix + dirDelim + suffix
           : suffix;
}
```

kafka-connect-s3/src/test/java/io/confluent/connect/s3/util/FileUtils.java
```diff
@@ -16,14 +16,19 @@
...
 import org.apache.kafka.common.TopicPartition;
 
 public class FileUtils {
   public static final String TEST_FILE_DELIM = "#";
   public static final String TEST_DIRECTORY_DELIM = "_";
 
   public static String fileKey(String topicsPrefix, String keyPrefix, String name) {
-    return topicsPrefix + TEST_DIRECTORY_DELIM + keyPrefix + TEST_DIRECTORY_DELIM + name;
+    String base = "";
```
kkonstantine: Similar to the code above?

klamouri:
Hi!

Thanks for the feedback I updated the PR accordingly :)

kkonstantine:
ok to test

PR #104, opened 2017-10-07
Fix new client creation to cater for URLs and region together

nitisht:
To connect to an S3 compatible service, connector should be able to accept both URL and region together.

Refer https://github.com/minio/minio/issues/3978 for example


nitisht:
[clabot:check]

kkonstantine:
ok to test

nitisht:
Great! Thanks @kkonstantine

PR #106, opened 2017-10-11
Add s3.sse.kms.key.id to specify a particular key in AWS KMS 

kkonstantine:


kkonstantine:
https://github.com/confluentinc/kafka-connect-storage-cloud/pull/91 was closed by accident while rebasing. Reopening here, @AlexPiermatteo let me know if you'd like to continue working on this one.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -188,6 +191,10 @@ private MultipartUpload newMultipartUpload() throws IOException {
         newObjectMetadata()
     ).withCannedACL(cannedAcl);
 
+    if (ssea.equals("aws:kms") && StringUtils.isNotBlank(awsKmsKeyId)) {
```
jklukas: This should use `SSEAlgorithm.KMS.getAlgorithm()` rather than the literal string "aws:kms"

jklukas:
It would be really nice to have valid values for SSEA show up in the docs (see s3.region in https://docs.confluent.io/current/connect/connect-storage-cloud/kafka-connect-s3/docs/configuration_options.html#s3-configuration-options for an example).

I believe that requires creating a validator and passing that to the configdef when defining  the option.

kkonstantine:
Note: At this point, before any commits are added, this PR is exactly at the state that was issued initially by @AlexPiermatteo

Comments are welcome. I'll also wait to see if @AlexPiermatteo would like to continue on this one.

andybritz:
Has there been any moment on this PR recently, we are fairly keen to have this capability added to the core project. Currently we use a customised build with this patch applied but would prefer to stick to the default version.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -332,6 +335,18 @@ public static ConfigDef newConfigDef() {
           Width.LONG,
           "S3 Proxy Password"
       );
+
+      configDef.define(
+          SSE_AWS_KMS_ID_CONFIG,
+          Type.STRING,
+          SSE_AWS_KMS_ID_DEFAULT,
+          Importance.LOW,
+          "The S3 Server Side Encryption KMS custom id/alias name.",
```
rhauch: Have you thought about using an enumeration so that the configdef describes the allowable values?

Also, it would probably be more beneficial to have a more clear description for those that don't know offhand what this entails, such as "S3 Server Side Encryption Key" for the title and for the description something like: "The name of the AWS Key Management Service (AWS-KMS) key to be used for server side encryption of the S3 objects. No encryption is used when no key is provided, but it is enabled when `aws:kms` is specified."

Also, when adding or changing any configuration properties we should also update the `kafka-connect-s3/docs/configuration_options.rst` file in this PR.

kkonstantine:
@jklukas @rhauch I have addressed your most recent comments and this PR seems ready to me. LMK what you think. Thanks!

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -63,6 +65,9 @@
   public static final String SSEA_CONFIG = "s3.ssea.name";
   public static final String SSEA_DEFAULT = "";
 
+  public static final String SSE_AWS_KMS_ID_CONFIG = "s3.sse.aws.kms.id";
+  public static final String SSE_AWS_KMS_ID_DEFAULT = "";
```
jklukas: The "aws" part of this configuration option seems redundant. "s3.sse.kms.id" (and renaming constants and methods here to match) would be more elegant. Not a blocker.

kkonstantine: Good point. Though you one can see 'aws' included in headers, it's better if we follow naming conventions in APIs such the one in `aws-cli`. 

Going with `s3.sse.kms.key.id`

kkonstantine:
Thanks for your reviews @rhauch @jklukas !

PR #108, opened 2017-10-20
HOTFIX: Handle primitive types in AvroFormat

kkonstantine:
* Follows from: https://github.com/confluentinc/kafka-connect-hdfs/issues/130
 * Equivalent to: https://github.com/confluentinc/kafka-connect-hdfs/pull/176

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -78,10 +78,9 @@ public class AvroRecordWriterProvider implements RecordWriterProvider<S3SinkConn
           // AvroData wraps primitive types so their schema can be included. We need to unwrap
           // NonRecordContainers to just their value to properly handle these types
           if (value instanceof NonRecordContainer) {
-            writer.append(((NonRecordContainer) value).getValue());
-          } else {
-            writer.append(value);
+            value = ((NonRecordContainer) value).getValue();
           }
+          writer.append(value);
         } catch (IOException e) {
           throw new ConnectException(e);
         }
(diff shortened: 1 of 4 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
```diff
@@ -74,7 +75,13 @@ public void write(SinkRecord record) {
...
         Object value = avroData.fromConnectData(schema, record.value());
         try {
-          writer.append(value);
+          // AvroData wraps primitive types so their schema can be included. We need to unwrap
+          // NonRecordContainers to just their value to properly handle these types
+          if (value instanceof NonRecordContainer) {
+            writer.append(((NonRecordContainer) value).getValue());
+          } else {
+            writer.append(value);
+          }
```
rhauch: Would it make more sense to do something like this inside the try block? (I've removed comments to keep things simple.)

```
if (value instanceof NonRecordContainer) {
    // put comment here
    value = ((NonRecordContainer) value).getValue();
}
writer.append(value);
```

rhauch: Just to clarify, this wrapping in a `NonRecordContainer` only happens at the top-level value in a record?

rhauch: BTW, I do see that this is the same code that was added in https://github.com/confluentinc/kafka-connect-hdfs/pull/176, but still seems like that is a bit more complicated.

kkonstantine: Actually, that's how I wrote it in the test I'm writing now. I like that more too.

kkonstantine: And also, yes, my understanding is that there's no nested application of `NonRecordContainer`.

PR #116, opened 2017-10-30
HOTFIX: Add recommender for schema generator classes only on 3.3.x.

kkonstantine:


rhauch:
@kkonstantine, are there any other recommenders that we should provide?

kkonstantine:
There's only `CREDENTIALS_PROVIDER_CLASS_CONFIG` but it has a default value. I decided to address this in another release.

kkonstantine:
ok to test

PR #118, opened 2017-11-03
MINOR: Regenerate docs and maintain order of configuration groups

rhauch:


kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -557,22 +557,25 @@ public String toString() {
   }
 
   public static ConfigDef getConfig() {
-    Map<String, ConfigDef.ConfigKey> everything = new HashMap<>(newConfigDef().configKeys());
```
kkonstantine: Probably using `LinkedHashMap` would suffice. Still, the solution is equivalent.

rhauch: It might be, except the order is still wrong since the overridden configs are added to `visible` first, and group order depends on insertion order. In order to maintain the order, it's easier to just use ConfigDef.

PR #128, opened 2017-11-27
[WIP] Implement Google Cloud Storage (GCS) sink connector.

kkonstantine:


kkonstantine:
This is initially submitted as PoC. It currently includes a lot of duplicated code with `kafka-connect-s3`. Before it gets merged, significant parts of it should be first fleshed out and included in a `cloud-common` module in the current repo that will host the common parts of all the cloud connectors in this repo (for example, the format package belongs in such a common module). Then this PR will be rebased on top of the `cloud-common` changes. However, refactoring the S3 connector on top the `cloud-common` doesn't have to happen before this PR is merged.

dsdinter:
I guess one interim approach to support GCS will be to reuse same library and methods but replacing the below:

1. Change the request endpoint to use the Cloud Storage request endpoint.
2. Replace the Amazon Web Services (AWS) access and secret key with the corresponding Cloud Storage access key and secret key (collectively called your Google developer key).

https://cloud.google.com/storage/docs/migrating#migration-simple

farmdawgnation:
Just curious, is there any ETA for this work landing in the master branch?

We're interested in using GCS with this connector, but we're going to need to disable multipart uploads to do so as it seems the interoperability API doesn't support multipart. Which... seems less than ideal?

msample:
As a workaround to use this this S3 connect with GCS, one can use Minio as an S3 API gateway to GCS native API.

msample:
As a workaround to use this this S3 connector with GCS, one can use Minio as an S3 API gateway to GCS native API. Worked for me in a test environment.

minio gateway gcs --address localhost:9000 <goog-proj-id>

connector has: store.url=http://localhost:9000

details: https://github.com/minio/minio/blob/master/docs/gateway/gcs.md

alex-shpak:
Hi @msample!
I also trying to setup minio as gateway, and I got a question,
how did you set AWS endpoint for sink to point at minio host?

Thanks!

msample:
@alex-shpak if you're sure Minio is up and running on port 9000, I'd make sure your store.url is on the connector (not the worker properties).  Properties like the following should get your connector going with Confluent 4.x

    "topics.regex":  (your db tables here)
    "s3.bucket.name": "your-bucket",
    "s3.part.size": "5242880",
    "storage.class": "io.confluent.connect.s3.storage.S3Storage",
    "path.format": "YYYY/MM/dd/HH",
    "store.url": "http://localhost:9000",

Note that the Minio GCS gateway only supports using credentials of one Google acct so you need the bucket owners to grant read & write access to that one Google acct (best way to manage it).  IIRC Minio stores some state in GCS for multipart uploads so it needs read access to restore that state if it's restarted.

alex-shpak:
Thank you for help,
my issue was that I omitted `http://` in minio url, implying that it's http by default (not https) which is not

kkonstantine:
Closing this PR since this work has been released and is in GA version here: 

https://www.confluent.io/connector/kafka-connect-gcs/

Cheers!

PR #138, opened 2018-02-01
CC-1490: NPE when rotate.interval.ms is set but partitioner is not instance of TimeBasedPartitioner

kkonstantine:


kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -114,6 +114,15 @@ public TopicPartitionWriter(TopicPartition tp,
     flushSize = connectorConfig.getInt(S3SinkConnectorConfig.FLUSH_SIZE_CONFIG);
     topicsDir = connectorConfig.getString(StorageCommonConfig.TOPICS_DIR_CONFIG);
     rotateIntervalMs = connectorConfig.getLong(S3SinkConnectorConfig.ROTATE_INTERVAL_MS_CONFIG);
+    if (rotateIntervalMs > 0) {
+      log.warn(
+          "Property '{}' is set to '{}ms' but partitioner is not an instance of {}. This property"
+              + " is ignored.",
+          S3SinkConnectorConfig.ROTATE_INTERVAL_MS_CONFIG,
+          rotateIntervalMs,
+          TimeBasedPartitioner.class.getName()
```
wicknicks: how do we know that `partitioner` is not an instance of `TimeBasedPartitioner ` here?

kkonstantine: Good call. I meant to use the `timestampExtractor` initialized above, but I didn't.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -650,6 +650,40 @@ public void testNoFilesWrittenWithoutCommit() throws Exception {
...
+        String.valueOf(TimeUnit.MINUTES.toMillis(1))
+    );
+    setUp();
+
+    // Define the partitioner
+    Partitioner<FieldSchema> partitioner = new DefaultPartitioner<>();
+    partitioner.configure(parsedConfig);
+    TopicPartitionWriter topicPartitionWriter = new TopicPartitionWriter(
+        TOPIC_PARTITION, storage, writerProvider, partitioner,  connectorConfig, context);
+
```
wicknicks: this block of code is repeated a bunch of times. should we put it in a `@Before` method? And have `topicPartitionWriter` as a field in this class?

kkonstantine: Not sure which block are you referring to, but everything below `setUp` varies more or less between test cases (the partitioner or the records constructed). More importantly though I'd like to keep this fix as focused as possible because it goes all the way back to `3.2.x`

Thanks for the reviews!

wicknicks: I meant this block:

```
    // Define the partitioner
    Partitioner<FieldSchema> partitioner = new DefaultPartitioner<>();
    partitioner.configure(parsedConfig);
    TopicPartitionWriter topicPartitionWriter = new TopicPartitionWriter(
        TOPIC_PARTITION, storage, writerProvider, partitioner,  connectorConfig, context);
```

I did a quick check, and this block is duplicated in 3.2.x as well.

kkonstantine: I was referring to possible conflicts that will be generated. But the partitioner is not the same across test cases.

PR #140, opened 2018-02-02
Fix for BufferOverflowException in S3OutputStream

staslos:
There is two overloaded versions of `write` method on S3OutputStream. In situation when stream `buffer` becomes full by calling `public void write(byte[] b, int off, int len)`, and then `public void write(int b)` called for some reason, we'll get BufferOverflowException. To avoid it, we do check `!buffer.hasRemaining()` in both methods before returning and flush buffer if it's full.

Just to provide more context, I got this exception when extending `kafka-connect-s3` with custom format that uses LZOP compression before sending data to S3. In this case LZOP API finalizes compression of the chunk of data by writing a byte at the end of the S3 stream. If buffer of the stream is full by that time, I get BufferOverflowException. To be consistent, both methods now check if buffer is full before returning.

There is test that supports given idea.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
@@ -94,7 +94,7 @@ public class S3OutputStream extends OutputStream {
       return;
     }
 
-    if (buffer.remaining() < len) {
+    if (buffer.remaining() <= len) {
       int firstPart = buffer.remaining();
       buffer.put(b, off, firstPart);
       uploadPart();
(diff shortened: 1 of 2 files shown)


staslos:
I don't know how your Jenkins works but FYI, Kafka just bumped their snapshot version to 1.2.0-SNAPSHOT.
Build works fine locally with Kafka 1.1.0-SNAPSHOT available in dependencies

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -101,6 +101,9 @@ public void write(byte[] b, int off, int len) throws IOException {
       write(b, off + firstPart, len - firstPart);
     } else {
       buffer.put(b, off, len);
+      if (!buffer.hasRemaining()) {
```
kkonstantine: That's a great catch @staslos! 
However, I'd prefer to push a slightly different fix, to keep having a single path (as opposed to two) where upload can happen in this method. In the `if` branch I'd rewrite the condition as `if (buffer.remaining() <= len) {`

staslos:
@kkonstantine thank you for the suggestion. Please see my changes.

kkonstantine:
Merging to `4.0.x` with the test case. Will cherry-pick only the fix to backport it to `3.2.x` and `3.3.x`

PR #148, opened 2018-03-07
Rely on jackson.version from common pom instead of defining it ourselves

ewencp:


ewencp:
See https://github.com/confluentinc/common/pull/102 for where we're moving this to a more central location to make updating versions easier. This is as far back as we were relying on the common pom. Build won't pass until we get the other one merged and builds for 4.0.x common go through.

PR #160, opened 2018-04-10
CC-1793. Used full jitter backoff strategy from AWS SDK.

sansanichfb:
* New parameter `s3.retry.backoff.ms`(defaults to 500ms) has been added;
* Removed custom retry logic;
* Used full jitter predefined retry policy from AWS SDK;
* Updated docs and unit-tests.


kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -95,6 +95,21 @@ NO compatibility, BACKWARD compatibility, FORWARD compatibility and FULL compati
 
 Schema evolution in the S3 connector works in the same way as in the `HDFS connector <../../../connect-hdfs/docs/hdfs_connector.html#schema-evolution>`_.
   
+Automatic Retries
+-----------------
+Users' environments might experience network partitioning, reach AWS throttling limits etc.
+In order to have a robust system it makes sense to have a retry policy, however, it's important
+to avoid a thundering herd problem.
```
rhauch: Maybe something a bit closer to what's in ES docs:
> The S3 connector may experience problems writing to the S3 bucket, such as during network partitions, interruptions, or even AWS throttling limits. In many cases, the connector will retry the request a number of times before failing. To prevent from further overloading the network or S3 service, the connector uses an exponential backoff technique to give the network and/or service time to recover.  The technique adds randomness, called jitter, to the calculated backoff times to prevent a thundering herd, where large numbers of requests from many tasks are submitted concurrently and overwhelm the service. Randomness spreads out the retries from many tasks and should reduce the overall time required to complete all outstanding requests compared to simple exponential backoff. The goal is to spread out the requests to S3 as much as possible.

kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -95,6 +95,21 @@ NO compatibility, BACKWARD compatibility, FORWARD compatibility and FULL compati
...
+The S3 connector uses retry policy, predefined by ``com.amazonaws.retry.PredefinedRetryPolicies.SDKDefaultRetryCondition``
+and ``com.amazonaws.retry.PredefinedBackoffStrategies.FullJitterBackoffStrategy`` provided by AWS SDK.
```
rhauch: I liked your suggestion last night of not talking to much about the implementation. I'd remove this.

kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -95,6 +95,21 @@ NO compatibility, BACKWARD compatibility, FORWARD compatibility and FULL compati
...
+
+To calculate a maximum delay time before a next attempt, following formula ``${retry.backoff
+.ms} * 2 ^ (retry-1)`` is used.
```
rhauch: Maybe:
> The delay for retries is dependent upon the connector's ``retry.backoff.ms`` and ``s3.part.retries`` configuration properties. The actual delay is randomized, but the maximum delay can be calculated as a function of the number of retry attempts with ``${retry.backoff.ms} * 2 ^ (retry-1)``. In order to keep the maximum delay within a reasonable duration, it is capped at 24 hours.

How does `s3.part.retries` play into this?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -106,6 +107,15 @@
...
   public static final Password S3_PROXY_PASS_DEFAULT = new Password(null);
 
+  /**
+   * An arbitrary absolute maximum practical retry time.
+   */
+  public static final int MAX_RETRY_TIME_MS = (int) TimeUnit.HOURS.toMillis(24);
+  /**
+   * Maximum retry limit.
+   **/
+  public static final int MAX_RETRIES = 30;
```
rhauch: This is not mentioned in the docs.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -101,16 +106,45 @@ public ClientConfiguration newClientConfiguration(S3SinkConnectorConfig config)
...
+    PredefinedBackoffStrategies.FullJitterBackoffStrategy backoffStrategy = new
+        PredefinedBackoffStrategies.FullJitterBackoffStrategy(
+        config.getLong(RETRY_BACKOFF_CONFIG).intValue(),
+        MAX_RETRY_TIME_MS
+    );
+
+    RetryPolicy retryPolicy = new RetryPolicy(
+        PredefinedRetryPolicies.DEFAULT_RETRY_CONDITION,
+        backoffStrategy,
+        conf.getS3PartRetries(),
```
rhauch: Need to describe how the `s3.part.retries` property affects retry delays.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -101,16 +106,45 @@ public ClientConfiguration newClientConfiguration(S3SinkConnectorConfig config)
...
+  /**
+   * Creates a retry policy, based on full jitter backoff strategy
+   * and default retry condition.
+   *
+   * @param config the S3 configuration.
+   * @return retry policy
+   * @see com.amazonaws.retry.PredefinedRetryPolicies.SDKDefaultRetryCondition
+   * @see PredefinedBackoffStrategies.FullJitterBackoffStrategy
+   */
+  public RetryPolicy newRetryPolicy(S3SinkConnectorConfig config) {
```
rhauch: Should this be public? Maybe protected so it can be overridden in a subclass?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -101,16 +106,45 @@ public ClientConfiguration newClientConfiguration(S3SinkConnectorConfig config)
...
+          .withRetryPolicy(fullJitterPolicy);
     }
 
     return clientConfiguration;
   }
 
+
+  /**
+   * Creates a retry policy, based on full jitter backoff strategy
+   * and default retry condition.
```
rhauch: Add "This method currently configures the AWS client retry policy to use full jitter."

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
kkonstantine: I'd call this `fullJitterRetryPolicy` since it's not a generic `newRetryPolicy` method and as @rhauch said I'd made this `protected`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -126,13 +124,8 @@ private void uploadPart(final int size) throws IOException {
       multiPartUpload = newMultipartUpload();
     }
     try {
-      retry(new Runnable() {
-        public void run() {
-          multiPartUpload.uploadPart(new ByteArrayInputStream(buffer.array()), size);
-        }
-      }, retries, "Part upload failed");
```
kkonstantine: After removing this code, `retries` member variable is unused and needs to be removed too.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
kkonstantine: `MAX_RETRIES` is unused outside tests.

kkonstantine: What we use instead is the config option `S3_PART_RETRIES_CONFIG`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -101,16 +106,45 @@ public ClientConfiguration newClientConfiguration(S3SinkConnectorConfig config)
     clientConfiguration.withUserAgentPrefix(version);
     if (StringUtils.isNotBlank(config.getString(S3_PROXY_URL_CONFIG))) {
       S3ProxyConfig proxyConfig = new S3ProxyConfig(config);
+      RetryPolicy fullJitterPolicy = newRetryPolicy(config);
```
kkonstantine: nit: Could also be called in place in the `withRetryPolicy` especially if the method is renamed.

kafka-connect-s3/docs/s3_connector.rst
sansanichfb: ``s3.part.retries`` defines max number of attempts. 
The real number of attempts might be lesser since AWS client has some internal logic to determine whether it should retry request or not. One example - ``com.amazonaws.util.CapacityManager`` used to keep track of retries of failed requests(except throttle failures), so once capacity if drained - failed request won't be retried even if a number of attempts is lesser than ``s3.part.retries``.
Having all said above, I feel like we shouldn't dive that deep into internals but rather mention a high-level description of ``s3.part.retries``.

kafka-connect-s3/docs/s3_connector.rst
sansanichfb: Sure, will borrow this section as much as applicable to S3 connector.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
sansanichfb: Good catch, thanks, updated.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
sansanichfb: @rhauch There is no need to make it public, but to make unit-testing possible in isolated fashion, I made it public(it's also the same as other similar methods -`io.confluent.connect.s3.storage.S3Storage#newClientConfiguration`, `io.confluent.connect.s3.storage.S3Storage#newS3Client`). Since this method doesn't change the internal state of the `S3Storage` I thought it might be not a big harm.

@kkonstantine sure, renamed.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
sansanichfb: Do you mean to add this to the method `io.confluent.connect.s3.storage.S3Storage#newS3Client`?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
sansanichfb: Ok, will add explanation and example as in ES.

kafka-connect-s3/docs/s3_connector.rst
sansanichfb: Thanks, removed.

kafka-connect-s3/docs/s3_connector.rst
sansanichfb: ✅

kafka-connect-s3/docs/s3_connector.rst
sansanichfb: ✅

kafka-connect-s3/docs/configuration_options.rst
```diff
@@ -148,12 +148,19 @@ S3
...
 ``s3.part.retries``
-  Number of upload retries of a single S3 part. Zero means no retries.
+  Maximum number of retry attempts for failed requests. Zero means no retries. An actual number of attempts is determined by the S3 client based on multiple factors, including, but not limited to - the value of this parameter, type of exception occurred, a number of attempts to retry failed request(s) etc.
 
   * Type: int
   * Default: 3
   * Importance: medium
 
+``s3.retry.backoff.ms``
+  How long to wait in milliseconds before attempting the first retry of a failed S3 request. Upon a failure, this connector may wait up to twice as long as the previous wait, up to the maximum number of retries.This avoids retrying in a tight loop under failure scenarios.
```
sansanichfb: Borrowed description from ES connector.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -282,13 +291,32 @@ public static ConfigDef newConfigDef() {
...
 
+      configDef.define(
+          S3_RETRY_BACKOFF_CONFIG,
+          Type.LONG,
+          S3_RETRY_BACKOFF_DEFAULT,
+          Importance.LOW,
+          "How long to wait in milliseconds before attempting the first retry "
+              + "of a failed S3 request. Upon a failure, this connector may wait up to twice as "
+              + "long as the previous wait, up to the maximum number of retries."
+              + "This avoids retrying in a tight loop under failure scenarios.",
```
rhauch: Need a space after the period on line 312.

kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -95,6 +95,73 @@ NO compatibility, BACKWARD compatibility, FORWARD compatibility and FULL compati
...
+The delay for retries is dependent upon the connector's ``s3.retry.backoff.ms`` configuration
+property. The actual delay is randomized, but the maximum delay can be calculated as a function
+of the number of retry attempts with ``${s3.retry.backoff.ms} * 2 ^ (retry-1)``, where ``retry``
+is the number of attempts taken so far in the current iteration. In order to keep the maximum delay within a reasonable duration, it is capped at 24 hours.
+
+The maximum number of retry attempts is dictated by the ``s3.part.retries`` S3 connector configuration property, which defaults
+to 3 attempts. The backoff time, which is the amount of time to wait before retrying, is a function of the
+retry attempt number and the initial backoff time specified in the ``s3.retry.backoff.ms``
+connector configuration
+property, which defaults to 500 milliseconds. For example, the following table shows the possible wait times
```
rhauch: This last sentence is somewhat repetitive of the previous paragraph. Might be able to simply remove it, as long as you mention the default in the previous paragraph.

Or, maybe combine the two paragraphs:

> The maximum number of retry attempts is dictated by the ``s3.part.retries`` S3 connector configuration property, which defaults to 3 attempts. The delay for retries is dependent upon the connector's ``s3.retry.backoff.ms`` configuration property, which defaults to 500 milliseconds. The actual delay is randomized, but the maximum delay can be calculated as a function of the number of retry attempts with ``${s3.retry.backoff.ms} * 2 ^ (retry-1)``, where ``retry`` is the number of attempts taken so far in the current iteration. In order to keep the maximum delay within a reasonable duration, it is capped at 24 hours.
>
> For example, the following table ...

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -99,6 +110,8 @@ public ClientConfiguration newClientConfiguration(S3SinkConnectorConfig config)
 
     ClientConfiguration clientConfiguration = PredefinedClientConfigurations.defaultConfig();
     clientConfiguration.withUserAgentPrefix(version);
+    RetryPolicy fullJitterRetryPolicy = newFullJitterRetryPolicy(config);
+    clientConfiguration.withRetryPolicy(fullJitterRetryPolicy);
```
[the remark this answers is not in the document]
rhauch: Great suggestions, @wicknicks.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
rhauch: Renaming this method helps, and means my JavaDoc suggestion isn't really needed since the method name explains that it's using full jitter. So you can disregard this comment.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
rhauch: I like the additions. Thanks.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -111,6 +124,33 @@ public ClientConfiguration newClientConfiguration(S3SinkConnectorConfig config)
...
+  /**
+   * Creates a retry policy, based on full jitter backoff strategy
+   * and default retry condition.
+   *
+   * @param config the S3 configuration.
+   * @return retry policy
+   * @see com.amazonaws.retry.PredefinedRetryPolicies.SDKDefaultRetryCondition
+   * @see PredefinedBackoffStrategies.FullJitterBackoffStrategy
+   */
+  public RetryPolicy newFullJitterRetryPolicy(S3SinkConnectorConfig config) {
```
rhauch: Wasn't there a discussion about making this protected?

kafka-connect-s3/docs/s3_connector.rst
sansanichfb: Makes sense, updated.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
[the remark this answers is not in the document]
sansanichfb: Thanks @wicknicks , nice suggestions, will update.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
sansanichfb: Thanks, fixed.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
sansanichfb: I've added Javadocs to those methods just in case.

sansanichfb:
@wicknicks , @rhauch , @kkonstantine please let me know if you have more suggestions.

kafka-connect-s3/docs/configuration_options.rst
```diff
@@ -148,12 +148,19 @@ S3
   * Importance: low
 
 ``s3.part.retries``
-  Number of upload retries of a single S3 part. Zero means no retries.
+  Maximum number of retry attempts for failed requests. Zero means no retries. An actual number of attempts is determined by the S3 client based on multiple factors, including, but not limited to - the value of this parameter, type of exception occurred, a number of attempts to retry failed request(s) etc.
```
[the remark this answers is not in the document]
sansanichfb: Thanks, fixed articles.

By last statements, I wanted to emphasize that the actual number of max attempts used in the S3 client can be lesser than the value of this parameter since S3 client tracks capacity of already retried failed requests and never goes above it even if `s3.part.retries` is bigger.
As a reference - https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/http/AmazonHttpClient.java#L1477.

By default, the S3 client has the capacity to retry failed request(s) 100 times, so there is a chance that we will exhaust that capacity earlier, therefore, I don't want users to be surprised in that case.

kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -95,6 +95,80 @@ NO compatibility, BACKWARD compatibility, FORWARD compatibility and FULL compati
 
 Schema evolution in the S3 connector works in the same way as in the `HDFS connector <../../../connect-hdfs/docs/hdfs_connector.html#schema-evolution>`_.
   
+Automatic Retries
+-----------------
+The S3 connector may experience problems writing to the S3 bucket, such as during network partitions,
```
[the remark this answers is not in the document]
rhauch: +1

kafka-connect-s3/docs/configuration_options.rst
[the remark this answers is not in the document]
kkonstantine: I see. The sentence _a number of attempts to retry failed request(s) etc_ doesn't seem to make this clear. 
I'd suggest that we skip it, or say something like: 

_throttling settings of the underlying S3 client, etc_ 

since this is indeed controlled by the `com.amazonaws.sdk.enableThrottledRetry` property and what's used by default in the client we instantiate.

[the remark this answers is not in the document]
sansanichfb: Sure, let me reframe it a bit.

kafka-connect-s3/docs/s3_connector.rst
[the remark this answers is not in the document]
sansanichfb: Thanks, fixed.

kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -95,6 +95,80 @@ NO compatibility, BACKWARD compatibility, FORWARD compatibility and FULL compati
...
+The technique adds randomness, called jitter, to the calculated backoff times to prevent a thundering herd,
+where large numbers of requests from many tasks are submitted concurrently and overwhelm the service.
+Randomness spreads out the retries from many tasks and should reduce the overall time required
+to complete all outstanding requests compared to simple exponential backoff.
+The goal is to spread out the requests to S3 as much as possible.
+
+The maximum number of retry attempts is dictated by the ``s3.part.retries`` S3 connector
+configuration property,
+which defaults to 3 attempts. The delay for retries is dependent upon the connector's ``s3.retry.backoff.ms``
+configuration property, which defaults to 500 milliseconds. The actual delay is randomized,
```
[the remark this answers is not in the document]
sansanichfb: Fixed.

kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -95,6 +95,80 @@ NO compatibility, BACKWARD compatibility, FORWARD compatibility and FULL compati
...
+For example, the following table shows the possible wait times
+before submitting each of the 3 retry attempts:
+
+.. table:: Range of backoff times for each retry using the default configuration
+   :widths: auto
+
+   =====  =====================  =====================  ==============================================
+   Retry  Minimum Backoff (sec)  Maximum Backoff (sec)  Total Potential Delay from First Attempt (sec)
+   =====  =====================  =====================  ==============================================
+     1         0.0                      0.5                              0.5
```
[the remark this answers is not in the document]
sansanichfb: Fixed.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -111,6 +127,33 @@ public ClientConfiguration newClientConfiguration(S3SinkConnectorConfig config)
...
+   *
+   * @param config the S3 configuration.
+   * @return retry policy
+   * @see com.amazonaws.retry.PredefinedRetryPolicies.SDKDefaultRetryCondition
+   * @see PredefinedBackoffStrategies.FullJitterBackoffStrategy
+   */
+  protected RetryPolicy newFullJitterRetryPolicy(S3SinkConnectorConfig config) {
+
+    PredefinedBackoffStrategies.FullJitterBackoffStrategy backoffStrategy = new
+        PredefinedBackoffStrategies.FullJitterBackoffStrategy(
```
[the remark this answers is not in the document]
sansanichfb: Thanks, fixed.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -116,6 +119,14 @@
   public static final String S3_PROXY_PASS_CONFIG = "s3.proxy.password";
   public static final Password S3_PROXY_PASS_DEFAULT = new Password(null);
 
+  /**
+   * An arbitrary absolute maximum practical retry time.
+   */
+  public static final int S3_RETRY_MAX_TIME_MS = (int) TimeUnit.HOURS.toMillis(24);
+
+  public static final String S3_RETRY_BACKOFF_CONFIG = "s3.retry.backoff.ms";
+  public static final int S3_RETRY_BACKOFF_DEFAULT = 500;
```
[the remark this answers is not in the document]
sansanichfb: Updated to 200.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3ProxyTest.java
```diff
@@ -54,7 +72,18 @@
   @Before
   public void setUp() throws Exception {
     super.setUp();
-    storage = new S3Storage(connectorConfig, url, S3_TEST_BUCKET_NAME, null);
+    storage = new S3Storage(connectorConfig, url);
+    AmazonS3 s3 = Whitebox.getInternalState(storage, "s3");
```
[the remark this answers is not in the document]
sansanichfb: That isn't really a mocking internal state, but rather getting a private fields' states to assert their behavior.
In this case, I want to read the retry policy which has been used to configure and build a client. Since we agreed to make `io.confluent.connect.s3.storage.S3Storage#newFullJitterRetryPolicy` method protected, and extending the `S3Storage` class in this test is not the most elegant way I use `Whitebox` to read private fields here.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3ProxyTest.java
```diff
@@ -247,11 +261,114 @@ public void testHTTPSProtocolHostPortUserPassConfigs() throws Exception {
...
+  @Test
+  public void testRetryPolicyRetriableSkewException() throws Exception {
+    setUp();
+    AmazonServiceException e = new AmazonServiceException("Retriable exception");
+    e.setErrorCode("RequestExpired");
+    assertTrue(retryPolicy.getRetryCondition().shouldRetry(null, e, 1));
+  }
+
+  @Test(expected = ConfigException.class)
+  public void testRetryPolicyNegativeDelay() throws Exception {
```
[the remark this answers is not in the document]
sansanichfb: Good catch, removed.

kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -95,6 +95,80 @@ NO compatibility, BACKWARD compatibility, FORWARD compatibility and FULL compati
...
+
+.. table:: Range of backoff times when reaching the cap of 24 hours
+   :widths: auto
+
+   =====  =====================  =====================  ==============================================
+   Retry  Minimum Backoff (sec)  Maximum Backoff (sec)  Total Potential Delay from First Attempt (sec)
+   =====  =====================  =====================  ==============================================
+    17         0.0                  32768.0                           65535.5
+    18         0.0                  65536.0                           131071.5
+    19         0.0                  86400.0                           217471.5
```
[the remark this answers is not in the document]
sansanichfb: IMO it get's it straight, that even though delay time is limited to 24h, the S3 connector will continue retrying after reaching max delay.

kafka-connect-s3/docs/s3_connector.rst
```diff
@@ -95,6 +95,80 @@ NO compatibility, BACKWARD compatibility, FORWARD compatibility and FULL compati
...
+    20         0.0                  86400.0                           303871.5
+    21         0.0                  86400.0                           390271.5
+    22         0.0                  86400.0                           476671.5
+    23         0.0                  86400.0                           563071.5
+   =====  =====================  =====================  ==============================================
+
+It's not advised to set ``s3.part.retries`` too high since making more attempts after reaching a cap of 24 hours isn't practical.
+You can adjust both the ``s3.part.retries`` and ``s3.retry.backoff.ms`` connector configuration
+properties to achieve
+the desired backoff and retry characteristics.
```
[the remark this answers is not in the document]
sansanichfb: Fixed.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3ProxyTest.java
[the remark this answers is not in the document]
kkonstantine: That's what I meant by mocking. In general, including capturing and inspecting internal state too. 

I suggested `protected` because this also gives package private access. So if the test is in the right package (here `io.confluent.connect.s3.storage`) then you have access without inheriting the class. 

This package doesn't exist in tests and you'll have to create it if you go down that path. If that's all, it still sounds better than whitebox.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/storage/S3StorageTest.java
```diff
@@ -0,0 +1,172 @@
...
+        if (retryBackoffMs < 0) {
+          assertEquals(0, result);
+        } else {
+          assertTrue(result >= 0L);
+          assertTrue(result <= maxResult);
+        }
+      }
+    }
+  }
+}
```
wicknicks: minor: no newline at the end of the file.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3ProxyTest.java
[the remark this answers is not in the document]
sansanichfb: Yes, I like this suggestion - same-package access level will make it much cleaner. Also, after moving `S3ProxyTest` to `io.confluent.connect.s3.storage` all three methods(`newClientConfiguration`, `newFullJitterRetryPolicy`, `newS3Client`) can be protected.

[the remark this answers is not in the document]
kkonstantine: Great! Let's not change access scope to existing methods. Only the new one. Feel free to add all the needed unit tests in S3StorageTest (I assume) regardless of scope (public, or protected).

kafka-connect-s3/src/test/java/io/confluent/connect/s3/storage/S3StorageTest.java
sansanichfb: oops, fixed.

PR #163, opened 2018-04-23
MINOR: Add all parameters which decide the value of periodicRotation

wicknicks:
This is how the log message will look like now: 

```
[2018-04-23 10:57:41,935] TRACE Should apply periodic time-based rotation (rotateIntervalMs: '60000', baseRecordTimestamp: 'null', timestamp: 'null', encodedPartition: 'partition=12', currentEncodedPartition: 'partition=12')? false (io.confluent.connect.s3.TopicPartitionWriter:310)
```

Signed-off-by: Arjun Satish <arjun@confluent.io>

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -309,10 +309,12 @@ public class TopicPartitionWriter {
 
     log.trace(
         "Should apply periodic time-based rotation (rotateIntervalMs: '{}', baseRecordTimestamp: "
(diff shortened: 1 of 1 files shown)


cricket007:
As mentioned by @cotedm, `encodedPartition` is already logged on previous trace statement

wicknicks:
@cricket007 it's hard to put together logs when they are coming from different lines (esp. when there are multiple s3 connector tasks running in the worker). this will put all important the variables in one message.

cricket007:
@wicknicks That's a good point. I suppose it would help with the [MDC logging is added to Connect](https://issues.apache.org/jira/browse/KAFKA-3816)

wicknicks:
@cricket007 yes that's a new feature. not sure how far back we will backport it (if at all).

PR #164, opened 2018-04-23
Add trace logging for SinkTask#put & TopicPartitionWriter#rotateOnTime

cricket007:
Adding logging to help debug partitioning+record output issues. 

cc. @cotedm

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -357,10 +357,11 @@ public class TopicPartitionWriter {
 
     log.trace(
         "Should apply periodic time-based rotation (rotateIntervalMs: '{}', baseRecordTimestamp: "
-            + "'{}', timestamp: '{}')? {}",
+            + "'{}', timestamp: '{}', currentEncodedPartition: '{}')? {}",
         rotateIntervalMs,
         baseRecordTimestamp,
         recordTimestamp,
+        currentEncodedPartition,
         periodicRotation
     );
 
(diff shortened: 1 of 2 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -357,10 +357,11 @@ private boolean rotateOnTime(String encodedPartition, Long recordTimestamp, long
 
     log.trace(
         "Should apply periodic time-based rotation (rotateIntervalMs: '{}', baseRecordTimestamp: "
-            + "'{}', timestamp: '{}')? {}",
+            + "'{}', timestamp: '{}', currentEncodedPartition: '{}')? {}",
```
cricket007: `encodedPartition` is on the previous line. No need to include here, IMO.

cotedm: looking a bit closer I'm not actually sure what the [previous trace](https://github.com/cricket007/kafka-connect-storage-cloud/blob/53d19b65abf276294754eb98c173e8fc0897f6f8/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L352-L356) buys us at all. `recordCount` isn't really used in this method except to make sure it's > 0 at the beginning and we won't trace this anyway if it isn't. Maybe drop the previous trace and add `encodedPartition` here? I'm also inclined to think this message should show the conditions tested so you don't have to go back to code to see why this is important. @wicknicks any thoughts?

cricket007: Oops, didn't see this duplicates #163. Closing this

PR #171, opened 2018-05-26
MINOR: Remove MaxPermSize since we dropped support for Java 7

ijuma:


ijuma:
@kkonstantine, not sure if there are other connect repositories that need to be updated in the same way.

kkonstantine:
LGTM

PR #176, opened 2018-06-22
MINOR: Unblock tests on jdk9 and jdk10

kkonstantine:


kkonstantine:
Thanks @rhauch ! 
I had to push another commit for java10 and mockito. LMK if it still looks ok overall.

PR #179, opened 2018-06-26
use jdk 7 for 4.1.x branch

jarekr:


kkonstantine:
@jarekr during packaging we explicitly disable not only running, but also compiling the unit tests. There's a relevant switch that's enabled for a few packages. But the unit tests always required java 8 in this repo.

jarekr:
I don't need this after all, thanks for explaining what's going on in the other PR.

PR #181, opened 2018-07-02
MINOR: Update config options to include SSE-C

kkonstantine:


rhauch:
@kkonstantine aren't these generated?

kkonstantine:
@rhauch we still run main for this repo to generate the configs. That's why we didn't have them when the feature was merged. I agree we should move to auto-gen

rhauch:
@kkonstantine, I'm not referring to auto-gen. Did we just not manually generate them on a previous PR (in which case this PR is fine), or do we need to modify the ConfigDefs to reflect these changes (in which case we need another PR)?

PR #183, opened 2018-07-06
Set version of maven-site and maven-project-info-reports to ensure compatibility

rhauch:
This is the same fix that has been applied to common to fix maven-site incompatibilities. This repository does not use common as a parent POM, so the same changes need to be applied here.


rhauch:
Since this is the same fixes that have been applied to common and it fixes the build, I'll go ahead and merge.

PR #187, opened 2018-07-21
CC-2313 Handle late arriving records in storage cloud sink connectors

cyrusv:
We had issues where late-arriving records would cause rollups of s3 files when they were part of the previous partition. 

For example, consider an s3-sink configured for hourly rollups. Now, if we receive records from a previous partition ("late-arriving") then a separate rollup is triggered, so that the record does not appear in the wrong partition. Strictly speaking, this is expected behavior. However, there is demand to have an option to avoid these extra rollups in the event of a system with freqent late-arriving records causing pileup of numerous small files.

The proposed strategy is to add a new "strict" config option that lets users choose to put records from a previous partition in the current partition if it arrives late. By default, we will be in "strict" mode, preserving our current behavior. Some users will choose "nonstrict" option to have predictable, consistent file sizes, at the expense of occasionally having records in the wrong partition

I wanted to check in and get feedback before grinding out all of the TODOs:
TODOs:
- Integrate with GCS and possible HDFS
- Explore whether this behaves as expected for "scheduled" rotations
- Explore putting the config in the "common" repo

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -38,6 +38,7 @@ import java.util.Queue;
 import io.confluent.common.utils.SystemTime;
 import io.confluent.common.utils.Time;
 import io.confluent.connect.s3.storage.S3Storage;
+import io.confluent.connect.storage.StorageSinkConnectorConfig;
 import io.confluent.connect.storage.common.StorageCommonConfig;
 import io.confluent.connect.storage.format.RecordWriter;
 import io.confluent.connect.storage.format.RecordWriterProvider;
@@ -115,7 +116,7 @@ public class TopicPartitionWriter {
     flushSize = connectorConfig.getInt(S3SinkConnectorConfig.FLUSH_SIZE_CONFIG);
     topicsDir = connectorConfig.getString(StorageCommonConfig.TOPICS_DIR_CONFIG);
     appendLateData = connectorConfig.getBoolean(
-        S3SinkConnectorConfig.APPEND_LATE_DATA
+        StorageSinkConnectorConfig.APPEND_LATE_DATA
     );
     rotateIntervalMs = connectorConfig.getLong(S3SinkConnectorConfig.ROTATE_INTERVAL_MS_CONFIG);
     if (rotateIntervalMs > 0 && timestampExtractor == null) {
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -64,6 +64,7 @@ public class TopicPartitionWriter {
   private final SinkTaskContext context;
   private int recordCount;
   private final int flushSize;
+  private final boolean appendLateData;
   private final long rotateIntervalMs;
   private final long rotateScheduleIntervalMs;
   private long nextScheduledRotation;
@@ -113,6 +114,9 @@ public class TopicPartitionWriter {
                                   : null;
     flushSize = connectorConfig.getInt(S3SinkConnectorConfig.FLUSH_SIZE_CONFIG);
     topicsDir = connectorConfig.getString(StorageCommonConfig.TOPICS_DIR_CONFIG);
+    appendLateData = connectorConfig.getBoolean(
+        S3SinkConnectorConfig.APPEND_LATE_DATA
+    );
     rotateIntervalMs = connectorConfig.getLong(S3SinkConnectorConfig.ROTATE_INTERVAL_MS_CONFIG);
     if (rotateIntervalMs > 0 && timestampExtractor == null) {
       log.warn(
@@ -208,7 +212,9 @@ public class TopicPartitionWriter {
               setNextScheduledRotation();
               nextState();
             } else {
-              currentEncodedPartition = encodedPartition;
+              if (!appendLateData || currentEncodedPartition == null) {
+                currentEncodedPartition = encodedPartition;
+              }
               SinkRecord projectedRecord = compatibility.project(record, null, currentValueSchema);
               writeRecord(projectedRecord);
               buffer.poll();
@@ -293,13 +299,10 @@ public class TopicPartitionWriter {
     if (recordCount <= 0) {
       return false;
     }
-    // rotateIntervalMs > 0 implies timestampExtractor != null
-    boolean periodicRotation = rotateIntervalMs > 0
-        && timestampExtractor != null
-        && (
-        recordTimestamp - baseRecordTimestamp >= rotateIntervalMs
-            || !encodedPartition.equals(currentEncodedPartition)
-    );
+    boolean periodicRotationIsConfiged = rotateIntervalMs > 0 && timestampExtractor != null;
+    boolean newPartition = !appendLateData && !encodedPartition.equals(currentEncodedPartition);
+    boolean periodicRotation = periodicRotationIsConfiged &&
+        (recordTimestamp - baseRecordTimestamp >= rotateIntervalMs || newPartition);
 
     log.trace(
         "Checking rotation on time with recordCount '{}' and encodedPartition '{}'",
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -213,6 +213,8 @@ public class TopicPartitionWriter {
               setNextScheduledRotation();
               nextState();
             } else {
+              // If we are appending late data, then we only need to reset currentEncodedPartition
+              // the first time for the write buffer (e.g. when currentEncodedPartition is null)
               if (!appendLateData || currentEncodedPartition == null) {
                 currentEncodedPartition = encodedPartition;
               }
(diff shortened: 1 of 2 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -63,6 +63,18 @@
   public static final String REGION_CONFIG = "s3.region";
   public static final String REGION_DEFAULT = Regions.DEFAULT_REGION.getName();
 
+  public static final String AVRO_CODEC_CONFIG = "avro.codec";
+  public static final String AVRO_CODEC_DEFAULT = "null";
+
+  public static final String ROTATE_STRICT_PARTITIONING_CONFIG = "s3.rotate.partitioning.string";
+  public static final Boolean ROTATE_STRICT_PARTITIONING_DEFAULT = true;
+  public static final String ROTATE_STRICT_PARTITIONING_DOC = "If true, the partitions will " +
```
kkonstantine: I feel we can say what this config enables in a simpler way.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -113,6 +114,9 @@ public TopicPartitionWriter(TopicPartition tp,
                                   : null;
     flushSize = connectorConfig.getInt(S3SinkConnectorConfig.FLUSH_SIZE_CONFIG);
     topicsDir = connectorConfig.getString(StorageCommonConfig.TOPICS_DIR_CONFIG);
+    strictPartitioning = connectorConfig.getBoolean(
```
kkonstantine: `appendLateData` or something ...

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -294,12 +300,9 @@ private boolean rotateOnTime(String encodedPartition, Long recordTimestamp, long
       return false;
     }
     // rotateIntervalMs > 0 implies timestampExtractor != null
```
kkonstantine: comment is outdated

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -294,12 +300,9 @@ private boolean rotateOnTime(String encodedPartition, Long recordTimestamp, long
...
-    boolean periodicRotation = rotateIntervalMs > 0
-        && timestampExtractor != null
-        && (
-        recordTimestamp - baseRecordTimestamp >= rotateIntervalMs
-            || !encodedPartition.equals(currentEncodedPartition)
-    );
+    boolean periodicRotation = rotateIntervalMs > 0 && timestampExtractor != null;
```
kkonstantine: again `strictPartitioning` is counter-intuitive. It'll be better to signal the new feature as `true` and the old functionality as `false` for the append feature. 

In any case, splitting this boolean expression is correct. We might also need comments though re: what every variable is meant to mean

cyrusv:
@kkonstantine thanks for the comments -- I will fix up the varnames and docs. My open question is how this should affect GCS and HDFS (I think they are the only other two connectors that use "cloud-common"). If I move this config to cloud-common, then I'll have to add support to those other two connectors. Wanted to confirm before moving forward?

kkonstantine:
@cyrusv adding the config to storage common, when it's not required (has a default) is a backwards compatible operation. The connector can chose not to use the config in that case. We just might need to hide it (filter it out) when it returns the set of its config properties (statically). Furthermore, GCS will use it and HDFS might use it as well.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -646,6 +642,221 @@ public void testWriteRecordTimeBasedPartitionFieldTimestampHours() throws Except
...
+        topicsDir, earlyDirPrefix, TOPIC_PARTITION, 1, extension, ZERO_PAD_FMT
+    ));
+
+
+    Collections.sort(records, new Comparator<Struct>() {
+      @Override
+      public int compare(Struct o1, Struct o2) {
+        long ts1 = (long) o1.get("timestamp");
+        long ts2 = (long) o2.get("timestamp");
+        return Long.compare(ts1, ts2);
```
kkonstantine: I bet `return Long.compare((long) o1.get("timestamp"), (long) o2.get("timestamp"));` (not crazy about var names such as `ts1`, `ts2`, ...)
For the same reason, in comparators, and other generic methods with two arguments, better names for the args are:
`public int compare(Struct left, Struct right) {` or `public int compare(Struct lhs, Struct rhs) {`

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -646,6 +642,221 @@ public void testWriteRecordTimeBasedPartitionFieldTimestampHours() throws Except
...
+
+    TopicPartitionWriter topicPartitionWriter = new TopicPartitionWriter(
+        TOPIC_PARTITION, storage, writerProvider, partitioner, connectorConfig, context);
+
+    String key = "key";
+    Schema schema = createSchemaWithTimestampField();
+
+    DateTime first = new DateTime(2017, 3, 2, 10, 0, DateTimeZone.forID("America/Los_Angeles"));
+    // One record every 20 sec, puts 3 records every minute/rotate interval
+    long advanceMs = 20000;
```
kkonstantine: same as above

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -646,6 +642,221 @@ public void testWriteRecordTimeBasedPartitionFieldTimestampHours() throws Except
...
+    long timestampNow = first.getMillis();
```
kkonstantine: Interesting. No `t0` here? Nice!

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -646,6 +642,221 @@ public void testWriteRecordTimeBasedPartitionFieldTimestampHours() throws Except
...
+    // 20 sec
+    timestampNow += advanceMs;
+    records.add(createRecordWithTimestampField(schema, timestampNow));
+
+    // 40 sec
+    timestampNow += advanceMs;
+    records.add(createRecordWithTimestampField(schema, timestampNow));
+
+    // 30 sec: This should not flush
+    timestampNow -= 10000;
```
kkonstantine: same as above

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -646,6 +642,221 @@ public void testWriteRecordTimeBasedPartitionFieldTimestampHours() throws Except
...
+    records.add(createRecordWithTimestampField(schema, timestampNow));
+
+    // 120 sec: This should flush
+    timestampNow += 90000;
```
kkonstantine: same as above

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -208,7 +213,9 @@ public void write() {
               setNextScheduledRotation();
               nextState();
             } else {
-              currentEncodedPartition = encodedPartition;
+              if (!appendLateData || currentEncodedPartition == null) {
```
kkonstantine: Tbh, a comment wouldn't probably hurt here. Especially, since we weren't checking for `null` before

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -646,6 +642,221 @@ public void testWriteRecordTimeBasedPartitionFieldTimestampHours() throws Except
...
+    setUp();
+
+    // Define the partitioner
+    Partitioner<FieldSchema> partitioner = new HourlyPartitioner<>();
+    parsedConfig.put(PartitionerConfig.TIMESTAMP_EXTRACTOR_CLASS_CONFIG, "RecordField");
+    parsedConfig.put(PartitionerConfig.SCHEMA_GENERATOR_CLASS_CONFIG, TimeBasedSchemaGenerator.class);
+    partitioner.configure(parsedConfig);
+
+    TopicPartitionWriter topicPartitionWriter = new TopicPartitionWriter(
+        TOPIC_PARTITION, storage, writerProvider, partitioner, connectorConfig, context);
```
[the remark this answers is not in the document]
cyrusv: I'm following the style you already used throughout the file, and our checkstyle and Google autoformatter both say this is fine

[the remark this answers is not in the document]
cyrusv: In general, I'd like for us as a team to put more into auto-formatting and auto-linting. Or maybe I'm doing something wrong?

cyrusv:
@kkonstantine I addressed all your suggestions!

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -663,12 +875,12 @@ public void testNoFilesWrittenWithoutCommit() throws Exception {
...
     partitioner.configure(parsedConfig);
     TopicPartitionWriter topicPartitionWriter = new TopicPartitionWriter(
-        TOPIC_PARTITION, storage, writerProvider, partitioner,  connectorConfig, context);
+        TOPIC_PARTITION, storage, writerProvider, partitioner, connectorConfig, context);
 
     String key = "key";
     Schema schema = createSchema();
     List<Struct> records = createRecordBatches(schema, 3, 3);
-
+TimeUnit.SECONDS.toMillis(20);
```
rhauch: What is this line attempting to do?

cyrusv: Just some extra math to keep the CPU on its toes!

removed -- it was a stray copy-pasta

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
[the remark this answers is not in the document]
kkonstantine: nits are mostly optional. In intellij I use `New line after '(' == true` and `Place ')' on new line | true`. Automation is nice. Still some things will remain in peoples discretion, which sometimes is also nice

PR #205, opened 2018-09-19
Add late-append-data to docs

cyrusv:


kafka-connect-s3/docs/configuration_options.rst
```diff
@@ -61,6 +61,13 @@ Connector
   * Valid Values: [0,...]
   * Importance: low
 
+``append.late.data``
```
joel-hamill: how about (`marked` = `named`?):

```
Indicates how the connector manages late-arriving data that is intended 
for earlier partitions. A value of ``true`` means that the connector will 
append late-arriving data into the currently open partition. A value of ``false`` 
means that the connector will roll late-arriving data into files that are named as 
the partition and extraction time, even if this creates many small files. 
```

cyrusv: @joel-hamill good suggestions -- this will also require an  update here:

cyrusv: To clarify, we should only be pint-merging this up to 5.0.x, right? Then we need to move these docs changes over to docs repo?

joel-hamill: @cyrusv yes, good catch - actually since they are now located natively in the docs repo i think you can close this PR and instead make it here: https://github.com/confluentinc/docs.

PR #213, opened 2018-10-30
Update node label for current Jenkins labeling

ewencp:


Jenkinsfile
```diff
@@ -2,5 +2,5 @@
 common {
   slackChannel = '#connect-eng'
   upstreamProjects = 'confluentinc/kafka-connect-storage-common'
-  nodeLabel = 'openjdk8-xenial'
+  nodeLabel = 'docker-oraclejdk8'
```
maxzheng: If we should be using jdk8 instead of 7 for these older branches, then ok. You should re-target to 3.2.x though.

ewencp: oops, thought i had gotten to the earliest branch with a Jenkinsfile. retargeted to 3.2.x.

this is an unusual case -- until we switched to jdk8, this one needed override because the library we need for testing the s3 connector was built with jdk8. it was purely for the tests, so packaging and anything else that doesn't run tests could still work with jdk7 (and we just accepted the risk that we might mess up using jdk8 features and have to fix it when we detect it elsewhere).

PR #220, opened 2018-11-29
CC-3035: Upgrade the S3Mock library so the module builds on Java 8 and Java 11

rhauch:
Also upgrade the S3Mock library from 0.1.5 (on Scala 2.11) to 0.2.5 (on Scala 2.12), which incorporates a number of bug fixes. This is to take advantage of a new `S3Mock.shutdown()` method that properly closes the Akka framework and HTTP listener created in the S3Mock object. For some reason, these still-bound HTTP listeners are not a problem on Java 8, but with Java 11 they prevent the next test from binding to the address and cause the test to fail.

With these changes, the build passes on both Java 8 and Java 11.

### DIFF
--- a/kafka-connect-s3/pom.xml
+++ b/kafka-connect-s3/pom.xml
@@ -37,16 +37,9 @@
 
     <properties>
         <aws.version>1.11.86</aws.version>
-        <s3mock.version>0.1.5</s3mock.version>
+        <s3mock.version>0.2.5</s3mock.version>
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
     </properties>
-
-    <repositories>
-        <repository>
-            <id>findify</id>
-            <url>https://dl.bintray.com/findify/maven/</url>
-        </repository>
-    </repositories>
 
     <dependencyManagement>
         <dependencies>
@@ -67,7 +60,7 @@
         </dependency>
         <dependency>
             <groupId>io.findify</groupId>
-            <artifactId>s3mock_2.11</artifactId>
+            <artifactId>s3mock_2.12</artifactId>
             <version>${s3mock.version}</version>
             <scope>test</scope>
         </dependency>
--- a/kafka-connect-s3/src/test/java/io/confluent/connect/s3/TestWithMockedS3.java
+++ b/kafka-connect-s3/src/test/java/io/confluent/connect/s3/TestWithMockedS3.java
@@ -81,7 +81,7 @@ public class TestWithMockedS3 extends S3SinkConnectorTestBase {
   @Override
   public void tearDown() throws Exception {
     super.tearDown();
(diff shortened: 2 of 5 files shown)


kafka-connect-s3/pom.xml
```diff
@@ -37,17 +37,10 @@
...
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
     </properties>
 
-    <repositories>
-        <repository>
-            <id>findify</id>
-            <url>https://dl.bintray.com/findify/maven/</url>
-        </repository>
-    </repositories>
-
```
rhauch: The S3Mock library is now in Maven Central, so this is no longer needed.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TestWithMockedS3.java
```diff
@@ -81,7 +81,7 @@ public void setUp() throws Exception {
   @Override
   public void tearDown() throws Exception {
     super.tearDown();
-    s3mock.stop();
+    s3mock.shutdown(); // shutdown the Akka and HTTP frameworks to close all connections
```
rhauch: This is the one line that we need to fix the Java 11 issue, and `shutdown()` was introduced in 0.2.5. All other changes in this PR address behavioral changes in the newer S3Mock version.

kafka-connect-s3/pom.xml
```diff
@@ -67,7 +60,7 @@
         </dependency>
         <dependency>
             <groupId>io.findify</groupId>
-            <artifactId>s3mock_2.11</artifactId>
+            <artifactId>s3mock_2.12</artifactId>
```
rhauch: AK 2.1 uses Scala 2.11, whereas AK trunk (which our `master` currently corresponds to) uses Scala 2.12. So this should not be backported further back than `master`.

xvrl: using `${kafka.scala.version}` should make this portable across branches

kafka-connect-s3/src/test/resources/log4j.properties
```diff
@@ -21,3 +21,5 @@ log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c:%L)%n
 
 log4j.logger.org.apache.kafka=ERROR
+#log4j.logger.io.confluent.connect.s3=DEBUG
+#log4j.logger.io.findify.s3mock=DEBUG
```
xvrl: we can probably remove those lines from the artifacts that get shipped.

kkonstantine: These are properties are in the test resources. We don't ship those IIRC

PR #223, opened 2018-12-14
Update copyright to use CCL

maxzheng:


wicknicks:
@maxzheng what was the reason to close this PR?

maxzheng:
@wicknicks This is used to create a new `debian-5.1+` branch and not meant to be merged into `debian` as that will continue to be used for building old versions of CP.

PR #172, opened 2018-06-06
[WIP] Add Parquet Format

eakmanrq:
Still being actively developed. Currently passing some basic tests.

PR #241, opened 2019-04-01
Add Parquet Format

tony810430:
After reviewing the works on #172 and other tests in the repo, I thought the overall implementation is great and test coverage is sufficient.

In order to accelerate helping this feature contributed back, I created this PR based on my own branch.
It is more convenient if there is any change need to make after reviewing from others.

The works in the PR also include:
1. Rebase all commits to master in order to merge by fast-forward easily.
2. Provide a new config to set parquet format compression type.
3. Fix some minor issues, such as coding style and removing redundant code.
---
### Implementation Brief:

Introduce `ParquetFormat` and `ParquetRecordWriterProvider` to receive records and write them into parquet file by converting `SinkRecord` to avro data and writing through `AvroParquetWriter`.

The inner class `S3ParquetOutputFile` in `ParquetRecordWriterProvider` implements `OutputFile` which should be provided in order to build `AvroParquetWriter`. Instead of other formats warp output stream into writer or just uses that stream directly, `AvroParquetWriter` uses `OutputFile`'s method to create output stream, so the implementation passes filename and s3 storage object to `S3ParquetOutputFile`'s constructor and creates the output stream when it is needed.

For the `S3OutputStream`, its interface changed to `PositionOutputStream`, which should implemented a new method,  so that it could be accepted by `ParquetWriter`.

Because we can't control when or how to commit file through `AvroParquetWriter` and the only way to manually commit file is to close it, we wrapped `S3OutputStream` by `S3ParquetOutputStream` to make sure `S3OutputStream#commit()` must be called when `S3ParquetOutputStream#close()` is called. Even through we don't know that is trigged by commit or close, it's okey due to idempotent property of s3 sink connect.

The last modified file is `S3SinkConnectorConfig`. Added a new configuration for parquet compression type, such as `gzip`, `snappy` and etc. All supported compression type from parquet library could be configured. Since I took the values from parquet library directly, which were not matched to exist s3 compression type config, I chose to introduce a new config name to distinguish them.

Last but not least, all of the unit tests for parquet format implementations took those from avro format as reference, and I think those tests are sufficient. Because we removed dependency to hive, we also lost some dependencies for parquet. I added them back as few as possible.


tony810430:
@kkonstantine, @eakmanrq
Please help to review this PR. Thanks.

btw, this PR's tests should rely on https://github.com/confluentinc/kafka-connect-storage-common/pull/99 merged first, since it used a new API from parquet-mr 1.10.0.

tony810430:
@kkonstantine, @eakmanrq
Would you have some time to review this PR? Thanks.

eakmanrq:
Code itself looks good to me. @tony810430 Have you had a chance to test this out in a production environment? Curious what the performance looks like.

tony810430:
@eakmanrq Yes, I have already try this out in our production environment. However, my use case is just a small topic with only one partition and little QPS. I have been running it over one week and have verified the uploaded data are all correct. I didn't see any critical performance downgrade during these days, but as I said above, it's hard to see how the performance is.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
```diff
@@ -0,0 +1,146 @@
...
+        } catch (IOException e) {
+          e.printStackTrace();
+        }
+      }
+    };
+  }
+
+  private static class S3ParquetOutputFile implements OutputFile {
+    private S3Storage storage;
+    private String filename;
```
kkonstantine: what this `filename` represents. It's not clear without a closer look in the PR. Would you mind editing the initial comment to describe a bit the implementation? In the meantime I'll take a closer look in the next few days. Thanks!

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -312,6 +316,21 @@ public static ConfigDef newConfigDef() {
...
 
+      configDef.define(
+          PARQUET_COMPRESSION_TYPE_CONFIG,
+          Type.STRING,
+          PARQUET_COMPRESSION_TYPE_DEFAULT,
+          new ParquetCompressionTypeValidator(),
+          Importance.LOW,
+          "Compression type for file written to S3. "
+            + "Applied when using ParquetFormat. "
+            + "Available values: uncompressed, snappy, gzip, lzo, brotli, lz4, zstd.",
```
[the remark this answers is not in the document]
tony810430: It seems that there is no `Type.ENUM` in `ConfigDef.java`[1], but it is better to provide a `Recommender`. Have addressed that.

[1]https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/config/ConfigDef.java#L779

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
tony810430: Quick response and will update later with the overall description of the implementation. This filename came from `RecordWriterProvider#getRecordWriter(conf, filename)`. Unlike other format, parquet's writer didn't provide `OutputStream` to writer builder directly, instead it provided `OutputFile` and builder would get stream from its method. Here, the implementation just wrapped `s3Storage` and `filename` into this class and used them to create `OutputStream` when method was called, but it could also wrap a created stream instead. What do you think?

tony810430:
@kkonstantine, I have updated some descriptions of implementations in my initial comment. Not sure if those are what you meant and wanted. Please let me know if you need more information.

jocelyndrean:
I tried this PR today and everything worked perfectly :) Thanks @tony810430 for this amazing job !

tony810430:
@jocelyndrean Thanks for sharing your experience. It will be better if you have free time to help review this PR and we can make this feature merged for more users to benefit from it. =)

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorConfigTest.java
```diff
@@ -329,5 +330,31 @@ public void testS3RetryBackoffNegative() {
     connectorConfig = new S3SinkConnectorConfig(properties);
     connectorConfig.getLong(S3SinkConnectorConfig.S3_RETRY_BACKOFF_CONFIG);
   }
+  
+  @Test
+  public void testParquetCompressionTypeSupported() {
```
jocelyndrean: Should we add brotli ?

tony810430: You are right. Have addressed that.

tony810430:
@kkonstantine Do you have time to review this PR recently?

kkonstantine:
Thanks for the follow ups @tony810430 
This feature can't make it to the upcoming 5.3 because we were past the feature freeze (and now code freeze) when we started discussing it. I'll have cycles again soon, and hopefully we can merge after that on master, targeting CP 5.4.

tony810430:
@kkonstantine Thanks for letting me know your current release progress. Please ping me if you prepare to review this PR again.

ankitchiplunkar:
Hi, this is an important issue when will this be merged?

CesarPantoja:
Hey everyone. Is there a timeline to merge this? Quite important for us.

reddykundan:
When are you guys planning on merging this ?

jgrillo-r7:
Is this still on track for CP 5.4?

kkonstantine:
Yes! 
We are on track for 5.4. Currently on PTO for another week will catch up and follow up with my review soon after I return.

deltabeat:
Hi - do you have an update on this for 5.4 please?

pom.xml
```diff
@@ -89,6 +115,12 @@
             <artifactId>connect-json</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.parquet</groupId>
```
kkonstantine: we don't need this block, we get it from `storage-common`. Also we shouldn't mark it as `provided` because the standalone archive of the connector (the zip that is constructed by the `kafka-connect-maven-plugin`) will not have all it needs.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3ParquetOutputStream.java
```diff
@@ -0,0 +1,39 @@
...
+  public S3ParquetOutputStream(S3OutputStream s3OutputStream) {
+    super(s3OutputStream.getKey(), s3OutputStream.getConnectorConfig(), s3OutputStream.getS3());
+    committed = false;
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (!committed) {
+      committed = true;
+      super.commit();
```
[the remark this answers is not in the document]
tony810430: Not sure if I understood you correctly, but I tried to address what you suggested.

tony810430:
@kkonstantine Have already addressed comments.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -533,6 +560,15 @@ public CompressionType getCompressionType() {
     return CompressionType.forName(getString(COMPRESSION_TYPE_CONFIG));
   }
 
+  public CompressionCodecName getCompressionCodecName() {
+    String parquetCompressionType = getString(PARQUET_COMPRESSION_TYPE_CONFIG);
+    if (parquetCompressionType.equals(PARQUET_COMPRESSION_TYPE_DEFAULT)) {
```
kkonstantine: hmm, this applies only to `none` right?
But if (for some reason) we changed the default value for this config then this would break. 
I'd suggest using specifically `"none"` here if this applies only to `"none"`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -160,12 +168,16 @@ public void commit() throws IOException {
...
+      internalClose();
     }
   }
 
   @Override
   public void close() throws IOException {
+    internalClose();
+  }
+
+  private void internalClose() throws IOException {
```
kkonstantine: this seems like a tautology. why do we need the `internalClose`? Is it a leftover?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -179,6 +179,10 @@ public OutputStream create(String path, S3SinkConnectorConfig conf, boolean over
   }
 
   public S3OutputStream create(String path, boolean overwrite) {
+    return create(path, overwrite, false);
+  }
+
+  public S3OutputStream create(String path, boolean overwrite, boolean isParquetFormat) {
```
kkonstantine: I'm a bit torn about creating yet another `public` method. People actually use them and extend them, and then we need to deprecate the ones that are not useful after some time. 

Again the information we need to decide whether to use parquet is available in the config. 
How about using internally: 
`ParquetFormat.class.isAssignableFrom(this.conf.getClass(S3SinkConnectorConfig.FORMAT_CLASS_CONFIG))` in order to decide?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3ParquetOutputStream.java
```diff
@@ -0,0 +1,45 @@
...
+  public S3ParquetOutputStream(String key, S3SinkConnectorConfig conf, AmazonS3 s3) {
+    super(key, conf, s3);
+    commit = false;
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (commit) {
+      super.commit();
+    } else {
```
kkonstantine: yes, that's what I had in mind. Let's unset the `commit` flag right after we commit. This should run in the same thread, so no race condition w.r.t. resetting this flag.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3ParquetOutputStream.java
```diff
@@ -0,0 +1,45 @@
...
+      super.close();
+    }
+  }
+
+  public void setCommit() {
+    this.commit = true;
```
kkonstantine: nit: we usually don't use `this` outside constructors unless we have to.

kkonstantine:
Also, `DataWriterParquetTest.testProjectNoVersion` seems to fail the build job. Are you able to reproduce locally?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
tony810430: It is a better choice. I did that because I hadn't found how to access config for format.class then. Thanks for your suggestion.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
tony810430: It prevents from infinite recursion. `S3ParquetOutputStream` overrides `close()`, and calls `super.commit()` (i.e. `S3OutputStream#commit()`), but `S3OutputStream#commit()` will also call `close()`, which will refer to `S3ParquetOutputStream#close()`. Hence, the infinite recursion happened.

Unset `commit` flag before calling `super.commit()` can solve this problem as well, but I'm not sure which implementation is better. What do you think?

tony810430:
I'm not sure why `DataWriterParquetTest.testProjectNoVersion` failed. It worked on my machine. And I also notice that it worked in the 12th build as well, and there is no change between 12th and 13th build.

I'll try to build other confluent's dependencies locally with latest version, and check if I can reproduce locally.

tony810430:
I found that there were some discrepancies between `DataWriterParquetTest` and `DataWriterAvroTest`. I have made them consistent, but not sure I have understood the intent from origin author.

tony810430:
I have run this feature since 5.2.x released version. I see it has `hadoop-mapreduce-client-core` jar in its `lib/` folder, but I didn't try to remove it when I upgraded to 5.3.x. I have no idea whether it is necessary outside tests.

For unified compression types, it seems that the current `Validator`'s API lacks other configs' settings like `Recommender` has, otherwise we can define `dependents` for `format.class` and access it in `CompressionTypeValidator`. For now, I don't have any good solution to verify `format.class` in the parsing phase.

kkonstantine:
With my two last commits I'm setting the hadoop dependencies version to the latest bugfix (`3.2.1`) and I'm removing the config `s3.parquet.compression.type`. The current PR will be merged without compression support for parquet, but shortly and again targeting CP 5.4 `parquet.codec` similar to `avro.codec` will be added to storage-common.

yauhen-sobaleu:
Hi guys, is it possible to test Parquet Format with S3 Connector with confluentinc/cp-kafka-connect:5.4.0-beta1 docker image? Is it already included?

bstaudacher:
Very excited for this! Thank you for your hard work @tony810430.
What release will this be available in? Can't wait to try it out.

NathanNam:
It will be part of CP 5.4.

dude0001:
When I try to use this, I'm getting an error that a Hadoop class is not found. Is this a defect or am I missing a prereq or other configuration in my container maybe?

```
Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted. Error: Could not initialize class org.apache.hadoop.conf.Configuration   [org.apache.kafka.connect.runtime.WorkerSinkTask]
--
java.lang.NoClassDefFoundError: Could not initialize class org.apache.hadoop.conf.Configuration
at org.apache.parquet.hadoop.ParquetWriter$Builder.&lt;init&gt;(ParquetWriter.java:345)
at org.apache.parquet.avro.AvroParquetWriter$Builder.&lt;init&gt;(AvroParquetWriter.java:162)
at org.apache.parquet.avro.AvroParquetWriter$Builder.&lt;init&gt;(AvroParquetWriter.java:153)
at org.apache.parquet.avro.AvroParquetWriter.builder(AvroParquetWriter.java:43)
at io.confluent.connect.s3.format.parquet.ParquetRecordWriterProvider$1.write(ParquetRecordWriterProvider.java:79)
at io.confluent.connect.s3.format.KeyValueHeaderRecordWriterProvider$1.write(KeyValueHeaderRecordWriterProvider.java:105)
at io.confluent.connect.s3.TopicPartitionWriter.writeRecord(TopicPartitionWriter.java:532)
at io.confluent.connect.s3.TopicPartitionWriter.checkRotationOrAppend(TopicPartitionWriter.java:302)
at io.confluent.connect.s3.TopicPartitionWriter.executeState(TopicPartitionWriter.java:245)
at io.confluent.connect.s3.TopicPartitionWriter.write(TopicPartitionWriter.java:196)
at io.confluent.connect.s3.S3SinkTask.put(S3SinkTask.java:234)
at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:586)
at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:329)
at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:232)
at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:201)
at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:185)
at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:234)
at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
at java.base/java.lang.Thread.run(Thread.java:834)
```

PR #244, opened 2019-04-30
CC-4433c: Remove unused dependency on Hive

kkonstantine:
• Hive dependencies need to be removed from Confluent platform packaging and S3 connector does not actually use any of the Hive integration hooks. This PR removes all the classes and interfaces that bring in Hive or Hadoop dependencies. 

• Partitioners are unaffected by the removal of FieldSchema as parameter, because the constructor of SchemaGenerator is never called from within the S3 connector.


kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorConfigTest.java
```diff
@@ -156,7 +155,7 @@ public String generatePartitionedPath(String topic, String encodedPartition) {
       }
 
       @Override
-      public List<FieldSchema> partitionFields() {
+      public List<Object> partitionFields() {
```
rhauch: Since this method should not be used in the S3 connector, would it be better to throw UnsupportedOperationException just to help reinforce that fact?

pom.xml
```diff
@@ -100,10 +100,6 @@
             <groupId>io.confluent</groupId>
             <artifactId>kafka-connect-storage-format</artifactId>
         </dependency>
-        <dependency>
-            <groupId>io.confluent</groupId>
-            <artifactId>kafka-connect-storage-hive</artifactId>
-        </dependency>
```
rhauch: Is the intention to remove this from the build but not remove the code? Why not just remove the module altogether?

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorConfigTest.java
kkonstantine: Good point! Added. Tests remained unaffected locally, which is expected since this method is not used in the connector.

rhauch: Perfect.

rhauch:
@kkonstantine, should we also remove the reference to `kafka-connect-hive` in the [packaging assembly descriptor](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/master/kafka-connect-s3/src/assembly/package.xml#L66)?

rhauch:
One more question: the `licenses` folder also includes the licenses for all of the dependencies, including [Hive](https://github.com/confluentinc/kafka-connect-storage-common/tree/master/licenses). Don't we want to include `license` folders under each of the modules, so that dependent projects get the correct licenses?

rhauch:
@kkonstantine, we probably also want to update the [`kafka-connect-s3/licenses` folder](https://github.com/confluentinc/kafka-connect-storage-cloud/tree/master/kafka-connect-s3/licenses), which currently includes Hive and Hadoop plus all the transitive dependencies.

kkonstantine:
Indeed. Removed licenses and notices that don't correspond to the dependencies any more.

PR #254, opened 2019-06-19
Added SafeByteArrayFormat to avoid using separator in S3SinkConnector

smallcampus:
**Background:**
We are using protobuf as our message schema in kafka. Messages are saved in kafka in byte array format.
When we try to backup our message with S3SinkConnector, we found that the ByteArrayFormat separate messages with separator. We hardly find a suitable separator for our messages.

**Solution:**
Therefore we try to follow the suggestion provided by google and create the SafeByteArrayFormat.
https://developers.google.com/protocol-buffers/docs/techniques?hl=en

The SafeByteArrayFormat write the size of each message before it write the message itself.
When we need to read back the messages from file, we read the first 4 bytes for the length of the message and then read (length of the message) bytes for the message body, repeat til the end.

**Limitation:**
For message size that is larger than Int value, there will be a problem..

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/SafeByteArrayRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/SafeByteArrayRecordWriterProvider.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2018 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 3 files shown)


phiSgr:
A signed int, with 31 bits, can represent a size of 2GB, thousands of times more than the default max message size. 

Alternatively the size can be a varint to save a few bytes for most messages, with no maximum value limitation. The cost is of course a bit of code complexity.

cjolivier01:
@smallcampus If this is still desired, can please resolve any conflicts?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/SafeByteArrayRecordWriterProvider.java
```diff
@@ -0,0 +1,94 @@
...
+
+      @Override
+      public void write(SinkRecord record) {
+        log.trace("Sink record: {}", record);
+        try {
+          byte[] bytes = converter.fromConnectData(record.topic(),
+                  record.valueSchema(), record.value());
+
+          ByteBuffer bb = ByteBuffer.allocate(4);
+          bb.putInt(bytes.length);
```
cjolivier01, line 70: why not long?

PR #271, opened 2019-09-12
Propagate exceptions thrown during time-based commits

### LINKED ISSUE #269: Failure to commit multipartupload results in apparent data loss
Hello,

We're running confluent-kafka-connect-s3.noarch version 5.1.2-1.
We have seen that when an attempt to commit a MultipartUpload fails, due to for example an S3 throttling error, this causes the task to just continue while leaving the S3 object that corresponds to the MultipartUpload invisible, since the commit is never done.
This seems to happen only when files are rotated on time.

Having the connector silently (except for log messages) continue after a failure like this can cause at least apparent data loss which reduces the confidence in the application. Perhaps this is by design, if so I would be glad to hear the rationale behind the decision. In my view, it would be better to fail the task completely in this scenario.

To walk through the flow as I understand it.
* An exception is thrown by the AWS lib and is caught here: https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.1.2-post/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java#L156
* This logs and rethrows the exception but in the finally clause, the buffer is closed and the stream itself is also marked as closed.
The rethrown exception is then caught here: https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.1.2-post/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L295 which then schedules a retry attempt.
* The retry attempt when it runs then reaches the commit method again but since the stream is closed, just silently returns: https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.1.2-post/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java#L146

In our logs we see these sequence of messages when it has happened:
* 2019-08-25 05:00:35,805 ERROR io.confluent.connect.s3.storage.S3OutputStream:157 - Multipart upload failed to complete for bucket 'bucketA' key 'some-prefix/objectA'
* 2019-08-25 05:30:34,605 WARN io.confluent.connect.s3.storage.S3OutputStream:141 - Tried to commit data for bucket 'bucketA' key 'some-prefix/objectA' on a closed stream. Ignoring. 
* 2019-08-25 05:30:34,605 DEBUG io.confluent.connect.s3.TopicPartitionWriter:484 - Committed some-prefix/objectA for topicA-2

We're not confident enough with this codebase to suggest a solution but would be very happy if someone has any ideas. 

Best Regards
Thomas

gharris1727:
ConnectExceptions that were thrown during time-based commits
were being logged but did not get retried. This causes data
loss because the framework assumes that such records were
committed successfully, when they were not. This fix throws
the error upwards, triggereing normal error handling.

Fixes #269

Signed-off-by: Greg Harris <gregh@confluent.io>

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -298,6 +298,7 @@ public class TopicPartitionWriter {
           log.error("Exception on topic partition {}: ", tp, e);
           failureTime = time.milliseconds();
           setRetryTimeout(timeoutMs);
+          throw e; // #269 propagate exceptions from time-based commits
         }
       }
 


aakashnshah:
Closed because I've added the fix and corresponding test to #274.

PR #280, opened 2019-10-17
CC-6766 Handle null-valued records in all three output types

cyrusv:
The new code will give consistent handling of null-valued records.

Before the change, json would correctly serialize null, while
bytes and avro would hard-fail. The recommended solution was to
use SMT to filter null-valued records, but this fix will integrate
the "skip" functionality into the connector. It does not preclude
the use of the SMT to get more general features such as DLQ.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -487,7 +487,7 @@ public class TopicPartitionWriter {
     ++recordCount;
   }
 
-  protected void commitFiles() {
+  private void commitFiles() {
     for (Map.Entry<String, String> entry : commitFiles.entrySet()) {
       commitFile(entry.getKey());
       log.debug("Committed {} for {}", entry.getValue(), tp);
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -487,7 +487,7 @@ public class TopicPartitionWriter {
     ++recordCount;
   }
 
-  private void commitFiles() {
+  protected void commitFiles() {
     for (Map.Entry<String, String> entry : commitFiles.entrySet()) {
       commitFile(entry.getKey());
       log.debug("Committed {} for {}", entry.getValue(), tp);
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -87,7 +87,8 @@ public class AvroRecordWriterProvider implements RecordWriterProvider<S3SinkConn
               log.debug("Null valued record cannot be written to output as Avro. "
                   + "Skipping. Record Key: {}", record.key());
               return;
-            } else if (conf.nullValueBehavior().equalsIgnoreCase(BehaviorOnNullValues.FAIL.toString())) {
+            } else if (conf.nullValueBehavior().equalsIgnoreCase(
+                BehaviorOnNullValues.FAIL.toString())) {
               throw new ConnectException("Null valued records are not writeable with current "
                   + S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG + " 'settings.");
             } else {
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -87,12 +87,9 @@ public class AvroRecordWriterProvider implements RecordWriterProvider<S3SinkConn
               log.debug("Null valued record cannot be written to output as Avro. "
                   + "Skipping. Record Key: {}", record.key());
               return;
-            } else if (conf.nullValueBehavior().equalsIgnoreCase(
-                BehaviorOnNullValues.FAIL.toString())) {
+            } else {
               throw new ConnectException("Null valued records are not writeable with current "
                   + S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG + " 'settings.");
-            } else {
-              throw new UnsupportedOperationException();
             }
           }
           writer.append(value);
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -81,6 +81,11 @@ public class AvroRecordWriterProvider implements RecordWriterProvider<S3SinkConn
           if (value instanceof NonRecordContainer) {
             value = ((NonRecordContainer) value).getValue();
           }
+          if (value == null) {
+            log.debug("Null valued record cannot be written out S3 output as Avro. "
+                + "Skipping. Record Key: {}", record.key());
+            return;
+          }
           writer.append(value);
         } catch (IOException e) {
           throw new ConnectException(e);
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.s3.format.avro;
 
+import io.confluent.connect.s3.S3SinkConnectorConfig.BehaviorOnNullValues;
 import org.apache.avro.file.CodecFactory;
 import org.apache.avro.file.DataFileWriter;
 import org.apache.avro.generic.GenericDatumWriter;
@@ -82,9 +83,16 @@ public class AvroRecordWriterProvider implements RecordWriterProvider<S3SinkConn
             value = ((NonRecordContainer) value).getValue();
           }
           if (value == null) {
-            log.debug("Null valued record cannot be written out S3 output as Avro. "
-                + "Skipping. Record Key: {}", record.key());
-            return;
+            if (conf.nullValueBehavior().equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
+              log.debug("Null valued record cannot be written to output as Avro. "
+                  + "Skipping. Record Key: {}", record.key());
+              return;
+            } else if (conf.nullValueBehavior().equalsIgnoreCase(BehaviorOnNullValues.FAIL.toString())) {
+              throw new ConnectException("Null valued records are not writeable with current "
+                  + S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG + " 'settings.");
+            } else {
+              throw new UnsupportedOperationException();
+            }
           }
           writer.append(value);
         } catch (IOException e) {
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
@@ -71,12 +71,9 @@ public class ByteArrayRecordWriterProvider implements RecordWriterProvider<S3Sin
               log.debug("Null valued record cannot be written to output as Avro. "
                   + "Skipping. Record Key: {}", record.key());
               return;
-            } else if (conf.nullValueBehavior()
-                .equalsIgnoreCase(BehaviorOnNullValues.FAIL.toString())) {
+            } else {
               throw new ConnectException("Null valued records are not writeable with current "
                   + S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG + " 'settings.");
-            } else {
-              throw new UnsupportedOperationException();
             }
           }
           s3outWrapper.write(bytes);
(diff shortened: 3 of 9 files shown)


kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterJsonTest.java
```diff
@@ -154,6 +154,37 @@ public void testGzipCompressionNoSchema() throws Exception {
...
+    List<String> fileNames = getExpectedFiles(new long[]{42L, 42L}, tp, ".json");
+    verifyFileListing(fileNames);
+    Collection<Object> records = readRecords(topicsDir,
+        getDirectory(tp.topic(), tp.partition()),
+        tp,
+        42,
+        ".json",
+        ZERO_PAD_FMT,
+        S3_TEST_BUCKET_NAME,
+        s3);
```
aakashnshah: formatting: put topicsDir on next line, and end parenthesis on its own line

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
```diff
@@ -165,6 +165,33 @@ public void testWriteRecordsOfEnumsWithEnhancedAvroData() throws Exception {
...
+    task = new S3SinkTask(connectorConfig, context, storage, partitioner, format, SYSTEM_TIME);
+
+    TopicPartition tp = context.assignment().iterator().next();
+    List<SinkRecord> sinkRecords = Collections.singletonList(new SinkRecord(TOPIC,
+        tp.partition(),
+        null,
+        "key",
+        null,
+        null,
+        42));
```
aakashnshah: formatting

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
```diff
@@ -165,6 +165,33 @@ public void testWriteRecordsOfEnumsWithEnhancedAvroData() throws Exception {
...
+    task.put(sinkRecords);
+    task.getTopicPartitionWriter(tp).commitFiles();
+    task.close(context.assignment());
+    task.stop();
+
+    List<String> fileNames = getExpectedFiles(new long[]{42L, 42L}, tp);
+    verifyFileListing(fileNames);
+    Collection<Object> records = readRecords(topicsDir, getDirectory(tp.topic(), tp.partition()), tp, 42,
+        extension, ZERO_PAD_FMT, S3_TEST_BUCKET_NAME, s3);
```
aakashnshah: formatting: everything on their own line

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterByteArrayTest.java
```diff
@@ -112,6 +114,34 @@ public void testNoSchema() throws Exception {
...
+  @Test
+  public void testNullValue() throws Exception {
+    localProps.put(S3SinkConnectorConfig.FORMAT_CLASS_CONFIG, ByteArrayFormat.class.getName());
+    setUp();
+    task = new S3SinkTask(connectorConfig, context, storage, partitioner, format, SYSTEM_TIME);
+
+    TopicPartition tp = context.assignment().iterator().next();
+    List<SinkRecord> sinkRecords = Collections.singletonList(
+        new SinkRecord(TOPIC, tp.partition(), null, "key",
+            null, null, 42));
```
aakashnshah: formatting

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterByteArrayTest.java
```diff
@@ -112,6 +114,34 @@ public void testNoSchema() throws Exception {
...
+    List<String> fileNames = getExpectedFiles(new long[]{42L, 42L}, tp, ".bin");
+    verifyFileListing(fileNames);
+    Collection<Object> records = readRecords(topicsDir,
+        getDirectory(tp.topic(), tp.partition()),
+        tp,
+        42,
+        ".bin",
+        ZERO_PAD_FMT,
+        S3_TEST_BUCKET_NAME,
+        s3);
```
aakashnshah: formatting

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterJsonTest.java
```diff
@@ -154,6 +154,37 @@ public void testGzipCompressionNoSchema() throws Exception {
...
+    task = new S3SinkTask(connectorConfig, context, storage, partitioner, format, SYSTEM_TIME);
+
+    TopicPartition tp = context.assignment().iterator().next();
+    List<SinkRecord> sinkRecords = Collections.singletonList(new SinkRecord(TOPIC,
+        tp.partition(),
+        null,
+        "key",
+        null,
+        null,
+        42));
```
aakashnshah: formatting

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
```diff
@@ -81,6 +82,19 @@ public void write(SinkRecord record) {
...
+            if (conf.nullValueBehavior().equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
+              log.debug("Null valued record cannot be written to output as Avro. "
+                  + "Skipping. Record Key: {}", record.key());
+              return;
+            } else if (conf.nullValueBehavior().equalsIgnoreCase(
+                BehaviorOnNullValues.FAIL.toString())) {
+              throw new ConnectException("Null valued records are not writeable with current "
+                  + S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG + " 'settings.");
+            } else {
+              throw new UnsupportedOperationException();
```
kkonstantine: That's confusing. This would be a bug that should be caught by a simple unit test. Throwing a different type of exception just for the sake of catching that bug in production doesn't seem useful to me. I suggest throwing the `ConnectException` for `FAIL`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
```diff
@@ -65,6 +66,19 @@ public void write(SinkRecord record) {
...
+            if (conf.nullValueBehavior().equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
+              log.debug("Null valued record cannot be written to output as Avro. "
+                  + "Skipping. Record Key: {}", record.key());
+              return;
+            } else if (conf.nullValueBehavior()
+                .equalsIgnoreCase(BehaviorOnNullValues.FAIL.toString())) {
+              throw new ConnectException("Null valued records are not writeable with current "
+                  + S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG + " 'settings.");
+            } else {
+              throw new UnsupportedOperationException();
```
kkonstantine: see above

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
```diff
@@ -80,6 +81,20 @@ public void write(SinkRecord record) {
...
+                    .equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
+                  log.debug("Null valued record cannot be written to output as Avro. "
+                      + "Skipping. Record Key: {}", record.key());
+                  return;
+                } else if (conf.nullValueBehavior()
+                    .equalsIgnoreCase(BehaviorOnNullValues.FAIL.toString())) {
+                  throw new ConnectException("Null valued records are not writeable with current "
+                      + S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG + " 'settings.");
+                } else {
+                  throw new UnsupportedOperationException();
```
kkonstantine: see above

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
```diff
@@ -165,6 +165,29 @@ public void testWriteRecordsOfEnumsWithEnhancedAvroData() throws Exception {
...
+    localProps.put(StorageSinkConnectorConfig.CONNECT_META_DATA_CONFIG, "true");
+    localProps.put(S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG, "ignore");
+    setUp();
+    task = new S3SinkTask(connectorConfig, context, storage, partitioner, format, SYSTEM_TIME);
+
+    TopicPartition tp = context.assignment().iterator().next();
+    List<SinkRecord> sinkRecords = Collections
+        .singletonList(new SinkRecord(TOPIC, tp.partition(), null, "key", null, null, 42));
+    task.put(sinkRecords);
+    task.getTopicPartitionWriter(tp).commitFiles();
```
kkonstantine: this hasn't been required so far in tests. We kept forcing commits by hitting the corresponding config (e.g. write 3 records with flush.size equal to 3). I suggest basing the test on previous ones without the requirement to change the scope to the method above.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterByteArrayTest.java
```diff
@@ -112,6 +113,28 @@ public void testNoSchema() throws Exception {
...
+    localProps.put(S3SinkConnectorConfig.FORMAT_CLASS_CONFIG, ByteArrayFormat.class.getName());
+    localProps.put(S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG, "ignore");
+    setUp();
+    task = new S3SinkTask(connectorConfig, context, storage, partitioner, format, SYSTEM_TIME);
+
+    TopicPartition tp = context.assignment().iterator().next();
+    List<SinkRecord> sinkRecords = Collections
+        .singletonList(new SinkRecord(TOPIC, tp.partition(), null, "key", null, null, 42));
+    task.put(sinkRecords);
+    task.getTopicPartitionWriter(tp).commitFiles();
```
kkonstantine: see above

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterJsonTest.java
```diff
@@ -154,6 +154,28 @@ public void testGzipCompressionNoSchema() throws Exception {
...
+    localProps.put(S3SinkConnectorConfig.FORMAT_CLASS_CONFIG, JsonFormat.class.getName());
+    localProps.put(S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG, "ignore");
+    setUp();
+    task = new S3SinkTask(connectorConfig, context, storage, partitioner, format, SYSTEM_TIME);
+
+    TopicPartition tp = context.assignment().iterator().next();
+    List<SinkRecord> sinkRecords = Collections
+        .singletonList(new SinkRecord(TOPIC, tp.partition(), null, "key", null, null, 42));
+    task.put(sinkRecords);
+    task.getTopicPartitionWriter(tp).commitFiles();
```
kkonstantine: see above

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -487,7 +487,7 @@ private void writeRecord(SinkRecord record) {
     ++recordCount;
   }
 
-  private void commitFiles() {
+  protected void commitFiles() {
```
kkonstantine: I don't think that's necessary. See my comments on tests.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
cyrusv: I think explicitly flushing greatly improves test legibility, and separates concerns out, lets this test focus on what it focuses on. I think the behavior of the other tests is very unobvious

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
cyrusv: I'm not following -- FAIL case already throws ConnectException.

Throwing UnsupportedOperationException (or any other RuntimeException) is probably the right thing to do here. Of course testing should catch these case, but assuming on bulletproof tests is not a good strategy. This is the same precedent already set in Elasticsearch connector (they actually throw RuntimeException, but I think Unsupported is more explicit). Throwing exception in production is absolutely useful because it stops progress if you are in illegal state

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
cyrusv: To illustrate the point, when I originally wrote the test, I saw no output files, which I thought might be expected behavior, then after going through a different case much later, I realized i should be seeing an empty file, not a missing file. Reading the test, it should be obvious what is happening, but a mystery flush size makes it quite challenging, and forces all the unit tests to conform or pass a ton of args around

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
kkonstantine: Probably I wasn't clear. My point was that `ConnectException` covers all the `else` cases, specialization doesn't seem useful, a) because this should never be executed, b) even if it did because of a bug it's more clear for the user to see a `ConnectException` rather than a `UnsupportedOperationException` that won't know what it means. Reducing boiler place might be preferable here.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
kkonstantine: I'm confused why you refer to flush a mystery. There are lots of tests exercising what `put` does, including timebased and size based rotations. You assert the result in several ways using the `verifyOffsets` and `verify` methods. I don't see why we can't follow the same pattern here

cyrusv:
@kkonstantine happy to add the test. I think in general we need to invest in organizing this code so we don't have to copy this 3 times.

cyrusv:
@kkonstantine I can't use `assertThrows` without upgrading the dependency, so I've opted to use `@Test(expected = ConnectException.class)`.

PR #281, opened 2019-10-23
CC-6801: Retry effectively on retriable failures while closing files

kkonstantine:
The connector so far has treated failures that occur when the output stream is closed as retriable only when `rotate.schedule.interval.ms` is set and it needs to flush outstanding records due to inactivity. 

However, offsets are not reset correctly, and retrying after the timeout expires might result in records being skipped as described in https://github.com/confluentinc/kafka-connect-storage-cloud/issues/269

This fix applies proper reset to the start offset that allows the consumer of a task to rewind correctly and considers any failure during file-commits (finalization of uploads) as a retriable exception.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -22,6 +22,7 @@ import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.errors.DataException;
+import org.apache.kafka.connect.errors.RetriableException;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.sink.SinkTask;
 import org.apache.kafka.connect.sink.SinkTaskContext;
@@ -51,6 +52,7 @@ public class S3SinkTask extends SinkTask {
 
   private S3SinkConnectorConfig connectorConfig;
   private String url;
+  private long timeoutMs;
   private S3Storage storage;
   private final Set<TopicPartition> assignment;
   private final Map<TopicPartition, TopicPartitionWriter> topicPartitionWriters;
@@ -93,6 +95,7 @@ public class S3SinkTask extends SinkTask {
     try {
       connectorConfig = new S3SinkConnectorConfig(props);
       url = connectorConfig.getString(StorageCommonConfig.STORE_URL_CONFIG);
+      timeoutMs = connectorConfig.getLong(S3SinkConnectorConfig.RETRY_BACKOFF_CONFIG);
 
       @SuppressWarnings("unchecked")
       Class<? extends S3Storage> storageClass =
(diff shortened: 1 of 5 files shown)


kkonstantine:
Initially, in this PR I'm targeting 5.0.x as the oldest release branch. I'll issue a subsequent PR without the Java 8 features to merge further back once we finalize the approach here.

kkonstantine:
Very accurate observations @gharris1727 and basically you described the approach which I followed and tested before I issue this PR in its current form. 

My conclusion is the following. Holding on to and resetting a complex object such as `TopicPartitionWriter` is much more error prone (now and in the future) than trying to clean up things and restart by issuing another TPW from scratch. The latter approach also reflects what happens if the task rebalances anyways. Also, there's no additional advantage on holding on to a TPW instance because the upload will restart as a whole anyways after you've discarded the s3 output stream. Given that only two methods are added, I don't think encapsulation is violated much anyways. Copy constructor seems it would inherit some of the disadvantages of resetting the object itself. 

Specifically, `failureTime` is exposed so that the changes in tracking the backoff are minimized. Not a very bad trade-off IMO. 

Finally the fewer the places that `context#offset` is called (ideally one) the better. The current bug remain undetected because exceptions are handled in two different places and because handling differed in the least frequent path. 

Let me know what you think.

gharris1727:
@kkonstantine I'm glad that you've thought about and tried out some of my suggestions. If you've found resetting the object to be too error-prone, i'm fine with discarding the object and starting anew.

However, I don't believe the copy constructor would inherit so many of the difficulties you've described. In my understanding, it would merely bring the duplicated constructor / helper method into the class, and expose fewer fields. The constructor would consist of copying the parameters passed into the initial TPW, and propagating the necessary details of the failure.

The error handling and `context::offset` call can remain centralized, with the benefits you've already listed.

kkonstantine:
@gharris1727 maybe I misunderstood the use of the term copy-constructor. Would you like to see a constructor that includes `failureTime` in its arguments? Because nothing is copied from the old object to the new one (that's how I read copy-constructor).

gharris1727:
I guess it wouldn't be a direct copy-constructor, but it would re-use all of the fields that were originally used to initialize the class, and perform some error-handling specific initialization:

```
// Constructor for retries on failure
TopicPartitionWriter(TopicPartitionWriter tpw) {
    this(tpw.tp,
           tpw.writerProvider,
           tpw.partitioner,
           tpw.connectorConfig,
           tpw.context,
           tpw.time
    );
    if (tpw.currentStartOffset != null) {
        context.offset(tp, tpw.currentStartOffset);
    }
    context.timeout(timeoutMs);
    this.failureTime = time.milliseconds();
}
```

kkonstantine:
I see. Yeah, I considered writing a static factory method on `TopicPartitionWriter`. But again, the only thing that would be added would be the `failureTime`. And since this is not a huge difference and since in theory it could make sense to set this after the creation of the object (and not always during its creation) I'd prefer to keep it as is. 

With respect to the other suggestions in your code patch @gharris1727 I'd like to keep this concerns in the task and not in the `TopicPartitionWriter`. For instance `currentStartOffset` is something that is returned but not set in the new instance (that's part of the reset logic). Also, context is better to be directly set in the task and not indirectly in TPW. 

Given the above, I agree that your suggestion is also applicable, but I'm inclined to keep the fix as is now since its equivalent and probably will result in fewer changes and conflicts when backported. Thanks for the detailed review. Lmk what you think.

rhauch:
@kkonstantine, do we need to backport this to branches earlier than `5.0.x`?

kkonstantine:
Thank you both! 
@rhauch yes, indeed, but I'd like to issue a separate PR and replace the java 8 specific code with java 7 code. Also, I'm anticipating that in previous release branches we might need a few additions that don't apply in more recent (e.g. post 5.0) branches.

kkonstantine:
(Will also wait for `5.4.x` to be created before merging. This is currently in process)

kkonstantine:
Squashed the PR to start working on backporting to `3.2.x`. Still waiting for `5.4.x` to be enabled before I merge here.

arogulin:
Thank you for fixing the issue! I see it is backported to 5.0.x and 5.1.x branches, but not 5.2.x. Do you know why would 5.2.x be ignored in this case?

PR #284, opened 2019-10-30
CC-6801: Retry effectively on retriable failures (backport to java 7 branches)

kkonstantine:
The connector so far has treated failures that occur when the output stream is closed as retriable only when rotate.schedule.interval.ms is set and it needs to flush outstanding records due to inactivity.

However, offsets are not reset correctly, and retrying after the timeout expires might result in records being skipped as described in #269

This fix applies proper reset to the start offset that allows the consumer of a task to rewind correctly and considers any failure during file-commits (finalization of uploads) as a retriable exception.


kkonstantine:
This is a simple backport of https://github.com/confluentinc/kafka-connect-storage-cloud/pull/281 in targeting release branches from `3.2.x` to `4.1.x`. These are java 7 branches. The changes are minimal. I'd appreciate a quick review.

Issue #201, opened 2018-09-12
Support for object-level "record count" in S3


danyanos:
Hi,

We were interested in contributing an enhancement that would add "record count" to all objects that land in S3. Traditionally our company performs a "row count" check when raw data enters a new datastore. The user landing the data provides an "expected" row count, and an internal framework compares the expected number of rows with the actual. Mismatches are used to alert producers of bad data so that some remediation steps can be taken. 

We've received requests from multiple users to add this "record count" to the objects that our Kafka Connect clusters are landing in S3. We've put together a proof of concept that uses S3 object-level tagging to add a "recordCount" tag to each object uploaded by the S3 Connector. Since we map each Kafka partition to one object in S3, we're also adding "startOffset" and "endOffset" tags to each object. We're interested in contributing this enhancement back to the community, but first wanted to check if this aligns with the roadmap for the S3 Connector and is something the community would support?

cricket007:
If anything, you could extend somewhere in here to add in the record count of the file to the S3 key. 

https://github.com/confluentinc/kafka-connect-storage-cloud/blob/master/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L473-L488

To not break existing expectations, you can add a configuration flag like `include.record.count`: `true/false`

kkonstantine:
Hi @danyanos 

this sounds definitely like a welcome contribution. I'd love to see a related PR. 
However, here, I'd like to discuss a bit whether addition record count as a tag vs as metadata header. 

To me, it seems that the latter might be slightly preferable for this type of information. For example, record count is associate with a specific object version and is immutable. A change in the count implies a different object version, as someone would expect. What do you think? Other than that, again I think it'll be useful, indeed. 

I wouldn't be in favor of adding this info to the key. Also, it doesn't need a config flag if we think it's useful and doesn't have side-effects.

danyanos:
Hey @kkonstantine, 

I can definitely clean up what we have and put together a PR. I saw there were some guidelines on how to build the project in the FAQ, but was curious if there was any guidance/best practices on submitting PRs? 

In terms of metadata vs tags this is something that we explored a bit. Unless there's something I'm missing it appears the AWS Java SDK only allows you to specify metadata when you initiate a multipart upload. Since we won't know the final record count until we complete the multipart upload this didn't seem like a viable option.

kevintdu:
Hi @kkonstantine

[...]

PR #291, opened 2019-11-16
S3 Object Level Tagging for Offsets & Record Counts

kevintdu:
PR for https://github.com/confluentinc/kafka-connect-storage-cloud/issues/201

This PR gives the S3 Connector Plugin the option to add s3 object level tags for starting offset, ending offset, and total record count of a given file. This should provide users a greater ability to track and monitor records written to s3.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -227,7 +227,7 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
           Type.BOOLEAN,
           S3_OBJECT_TAGGING_DEFAULT,
           Importance.LOW,
-          "Tag S3 objects with offsets & record count.",
+          "Tag S3 objects with start and end offsets, as well as record count.",
           group,
           ++orderInGroup,
           Width.LONG,
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -66,6 +66,9 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
   // S3 Group
   public static final String S3_BUCKET_CONFIG = "s3.bucket.name";
 
+  public static final String S3_OBJECT_TAGGING_CONFIG = "s3.object.tagging";
+  public static final boolean S3_OBJECT_TAGGING_DEFAULT = false;
+
   public static final String SSEA_CONFIG = "s3.ssea.name";
   public static final String SSEA_DEFAULT = "";
 
@@ -186,6 +189,18 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
           ++orderInGroup,
           Width.LONG,
           "S3 Bucket"
+      );
+
+      configDef.define(
+          S3_OBJECT_TAGGING_CONFIG,
+          Type.BOOLEAN,
+          S3_OBJECT_TAGGING_DEFAULT,
+          Importance.LOW,
+          "Tag S3 objects with offsets & record count.",
+          group,
+          ++orderInGroup,
+          Width.LONG,
+          "S3 Object Tagging"
       );
 
       configDef.define(
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -90,16 +90,6 @@ public class TopicPartitionWriter {
   private final S3SinkConnectorConfig connectorConfig;
   private static final Time SYSTEM_TIME = new SystemTime();
 
-  @Deprecated
-  public TopicPartitionWriter(TopicPartition tp,
-                              S3Storage storage,
-                              RecordWriterProvider<S3SinkConnectorConfig> writerProvider,
-                              Partitioner<?> partitioner,
-                              S3SinkConnectorConfig connectorConfig,
-                              SinkTaskContext context) {
-    this(tp, storage, writerProvider, partitioner, connectorConfig, context);
-  }
-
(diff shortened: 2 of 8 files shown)


kevintdu:
[clabot:check]

hunglin:
hi @kkonstantine, what do we need to do to merge this PR? If you are not the right person, could you point us in the right direction? Thanks in advance for your help.

kkonstantine:
hi @hunglin 
Thanks for the contribution! This can definitely be a useful feature. I'm tagging the Connect team at Confluent for review. This is a non-trivial feature, so we'll have to allocate some time to review. 

Given that we are in code freeze for the upcoming release (5.4.0) this PR correctly targets the `master` branch. Due to this timing, I'm hoping that we can get initial reviews within the next few weeks.

hunglin:
@kkonstantine thanks for the updates. Looking forward to seeing this new feature in the later release.

kevintdu:
Hi, wanted to check if code freeze is over for 5.4.0 release, and if there has been an opportunity to review this PR?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -218,6 +225,18 @@ public void delete(String name) {
...
   public void close() {}
 
+  public void addTags(String fileName, Map<String, String> tags)
+          throws SdkClientException {
+    List<Tag> newTagList = new ArrayList<>();
+    for (Map.Entry<String, String> entry : tags.entrySet()) {
+      newTagList.add(new Tag(entry.getKey(), entry.getValue()));
+    }
+    s3.setObjectTagging(new SetObjectTaggingRequest(
+            this.bucketName,
```
kkonstantine: nit: I also now realize that your tab size in some cases is equal to 4 spaces. 

In this project it should be: 
Tab size: 2
Indent: 2
Continuation indent: 4
(these are the settings in IntelliJ, similar should be set to other envs)

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -219,6 +222,18 @@ public static ConfigDef newConfigDef() {
           "S3 Bucket"
       );
 
+      configDef.define(
+          S3_OBJECT_TAGGING_CONFIG,
+          Type.BOOLEAN,
+          S3_OBJECT_TAGGING_DEFAULT,
+          Importance.LOW,
+          "Tag S3 objects with offsets & record count.",
```
kkonstantine: How about: 
"Tag S3 objects with start and end offsets, as well as record count"
cc @tobydrake7

tobydrake7: @kkonstantine looks good 👍

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -527,7 +545,32 @@ private void commitFile(String encodedPartition) {
...
+          storage.addTags(s3ObjectPath, tags);
+          log.info("Tagged file {} with record count {}", s3ObjectPath, recordCount);
+        } catch (SdkClientException e) {
+          log.trace("Unable to tag file {} with record count.", s3ObjectPath);
+        } catch (Exception e) {
+          log.trace("Unrecoverable exception while attempting to tag s3 object.");
+        }
+      }
+    }
     startOffsets.remove(encodedPartition);
```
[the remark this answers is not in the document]
kevintdu: Keeping the responsibilities separate makes sense. However, placing the call back to the initial position in commitFile() would remove the values before they are used for tagging, since commitFile() is called before tagging is called.

Went ahead and moved it to commitFiles() instead, after both commitFile() call and the optional tagFile() call has been complete.

kevintdu:
Updated based on comments

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -527,7 +545,34 @@ private void commitFile(String encodedPartition) {
...
+      tags.put("endOffset", Long.toString(endOffsets.get(encodedPartition)));
+      tags.put("recordCount", Long.toString(recordCounts.get(encodedPartition)));
 
-    startOffsets.remove(encodedPartition);
+      try {
+        storage.addTags(s3ObjectPath, tags);
+        log.info("Tagged file {} with starting offset {}, ending offset {}, record count {}",
+            s3ObjectPath, startOffsets, endOffsets, recordCount);
+      } catch (SdkClientException e) {
+        log.warn("Unable to tag file {}. Ignoring.", s3ObjectPath);
```
kkonstantine: We should add the exception here with the stacktrace. Swallowing the exception won't be very informative. 

I also wondered whether we should fail, and let it throw a retriable exception, but I think I agree with your decision to just skip tagging.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -527,7 +545,34 @@ private void commitFile(String encodedPartition) {
...
+      } catch (Exception e) {
+        log.warn("Unrecoverable exception while attempting to tag s3 object. Ignoring.");
```
kkonstantine: Let's include the exception here too.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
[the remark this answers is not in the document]
kkonstantine: Yeah, that's what I meant. Thanks!

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
kevintdu: Skip tagging works well with our internal use case.  It's something we can revisit based on user feedbacks.

kevintdu:
Exceptions & trace added to warn level logs. Thanks for the feedback.

PR #300, opened 2020-01-21
MINOR: Small fix to the config docs for S3 proxy

kkonstantine:


kkonstantine:
The build fails for an unrelated issue. Merging this minor doc change as-is

PR #302, opened 2020-02-04
CC-2539: Added a validator to check the format of S3 bucket name

PrajwalRavi:
Addressing Jira issue CC-2539: Added a validator to check the format of S3 bucket name.

### DIFF
--- a/kafka-connect-s3/pom.xml
+++ b/kafka-connect-s3/pom.xml
@@ -141,7 +141,7 @@
                             <goal>testCompile</goal>
                         </goals>
                         <configuration>
-                            <source>1.7</source>
+                            <source>1.8</source>
                             <target>1.8</target>
                         </configuration>
                     </execution>
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
@@ -18,8 +18,8 @@ package io.confluent.connect.s3;
 
 import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.ConfigValue;
 import org.apache.kafka.connect.connector.Task;
-import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.sink.SinkConnector;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -94,8 +94,14 @@ public class S3SinkConnector extends SinkConnector {
     Config config = super.validate(connectorConfigs);
     // Checking whether the bucket exists
     if (!S3BucketCheck.checkBucketExists(new S3SinkConnectorConfig(connectorConfigs))) {
-      throw new ConnectException("Bucket does not exist or invalid credentials "
-              + "provided");
+      for (ConfigValue configValue : config.configValues()) {
+        if (configValue.name().equals(S3SinkConnectorConfig.S3_BUCKET_CONFIG)) {
+          String bucketName = configValue.value().toString();
+          configValue.addErrorMessage(String.format("The S3 bucket named %s could not be found."
+                  + " Check the S3 credentials and verify the %s bucket exists.",
+              bucketName, bucketName));
+        }
+      }
     }
     return config;
   }
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
@@ -16,12 +16,8 @@
 
 package io.confluent.connect.s3;
 
-import io.confluent.connect.s3.storage.S3Storage;
-import io.confluent.connect.storage.StorageFactory;
-import io.confluent.connect.storage.common.StorageCommonConfig;
 import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigDef;
-import org.apache.kafka.common.config.ConfigValue;
 import org.apache.kafka.connect.connector.Task;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.sink.SinkConnector;
@@ -33,6 +29,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import io.confluent.connect.s3.util.S3BucketCheck;
 import io.confluent.connect.s3.util.Version;
 
 /**
@@ -92,36 +89,15 @@ public class S3SinkConnector extends SinkConnector {
     return S3SinkConnectorConfig.getConfig();
   }
 
-  private boolean checkBucketExists(S3SinkConnectorConfig config) {
-    @SuppressWarnings("unchecked")
-    Class<? extends S3Storage> storageClass =
-            (Class<? extends S3Storage>)
-                    config.getClass(StorageCommonConfig.STORAGE_CLASS_CONFIG);
-
-    final S3Storage bucketCheckHelper = StorageFactory.createStorage(storageClass,
-            S3SinkConnectorConfig.class, config,
-            config.getString(StorageCommonConfig.STORE_URL_CONFIG));
-
-    return bucketCheckHelper.bucketExists();
-  }
-
   @Override
   public Config validate(final Map<String, String> connectorConfigs) {
-    ConfigDef configDef = this.config();
-    if (null == configDef) {
-      throw new ConnectException(String.format("%s.config() must return a ConfigDef that is not"
-              + " null.", this.getClass().getName()));
-    } else {
-
-      // Checking whether the bucket exists
-      if (!checkBucketExists(new S3SinkConnectorConfig(connectorConfigs))) {
-        throw new ConnectException("Bucket does not exist or invalid credentials "
-                + "provided");
-      }
-
-      List<ConfigValue> configValues = configDef.validate(connectorConfigs);
-      return new Config(configValues);
+    Config config = super.validate(connectorConfigs);
+    // Checking whether the bucket exists
+    if (!S3BucketCheck.checkBucketExists(new S3SinkConnectorConfig(connectorConfigs))) {
+      throw new ConnectException("Bucket does not exist or invalid credentials "
+              + "provided");
     }
+    return config;
   }
 
 }
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
@@ -97,9 +97,13 @@ public class S3SinkConnector extends SinkConnector {
       for (ConfigValue configValue : config.configValues()) {
         if (configValue.name().equals(S3SinkConnectorConfig.S3_BUCKET_CONFIG)) {
           String bucketName = configValue.value().toString();
-          configValue.addErrorMessage(String.format("The S3 bucket named %s could not be found."
-                  + " Check the S3 credentials and verify the %s bucket exists.",
-              bucketName, bucketName));
+          configValue.addErrorMessage(
+              String.format("The S3 bucket named %s could not be found."
+                      + " Check the S3 credentials and verify the %s bucket exists.",
+                  bucketName,
+                  bucketName
+              )
+          );
         }
       }
     }
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
@@ -98,8 +98,9 @@ public class S3SinkConnector extends SinkConnector {
         if (configValue.name().equals(S3SinkConnectorConfig.S3_BUCKET_CONFIG)) {
           String bucketName = configValue.value().toString();
           configValue.addErrorMessage(
-              String.format("The S3 bucket named %s could not be found."
-                      + " Check the S3 credentials and verify the %s bucket exists.",
+              String.format(
+                  "The S3 bucket named '%s' could not be found. Check the S3 credentials and "
+                  + "verify the '%s' bucket exists.",
                   bucketName,
                   bucketName
               )
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
@@ -16,8 +16,14 @@
 
 package io.confluent.connect.s3;
 
+import io.confluent.connect.s3.storage.S3Storage;
+import io.confluent.connect.storage.StorageFactory;
+import io.confluent.connect.storage.common.StorageCommonConfig;
+import org.apache.kafka.common.config.Config;
 import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.ConfigValue;
 import org.apache.kafka.connect.connector.Task;
+import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.sink.SinkConnector;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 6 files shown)


kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterAvroTest.java
```diff
@@ -75,6 +79,9 @@
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
+@RunWith(PowerMockRunner.class)
```
PrajwalRavi: Required because this uses the class S3SinkConnectorTestBase which mocks a private static method.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorConfigTest.java
```diff
@@ -329,5 +340,25 @@ public void testS3RetryBackoffNegative() {
...
+  public void testBucketWithValidNameWhichExists() {
+    properties.put(S3SinkConnectorConfig.S3_BUCKET_CONFIG, "test-bucket");
+    connectorConfig = PowerMockito.spy(new S3SinkConnectorConfig(properties));
+  }
+
+  @Test(expected = ConnectException.class)
+  public void testBucketWithValidNameAndDoesNotExists() throws Exception {
+    properties.put(S3SinkConnectorConfig.S3_BUCKET_CONFIG, "test-bucket");
+    PowerMockito.doReturn(false).when(S3SinkConnectorConfig.class, "checkBucketExists",
+            Mockito.anyString());
```
aagarwal-gtr: change `Mockito.anyString()` to the actual bucket name its getting called with

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorTestBase.java
```diff
@@ -84,6 +85,9 @@ protected void starting(Description description) {
   @Override
   public void setUp() throws Exception {
     super.setUp();
+    PowerMockito.spy(S3SinkConnectorConfig.class);
+    PowerMockito.doReturn(true).when(S3SinkConnectorConfig.class, "checkBucketExists",
+            Mockito.anyString());
```
aagarwal-gtr: change `Mockito.anyString()` to the actual bucket name its getting called with

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -639,6 +645,33 @@ public String toString() {
...
+  private static boolean checkBucketExists(String bucketName) {
+    try {
+      final AmazonS3 bucketCheckHelper = AmazonS3ClientBuilder.defaultClient();
+      return bucketCheckHelper.doesBucketExist(bucketName);
+    } catch (SdkClientException e) {
+      return false;
+    }
+  }
+
+  private static class BucketValidator implements ConfigDef.Validator {
```
aagarwal-gtr: suggestion: `BucketNameValidator`?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -639,6 +645,33 @@ public String toString() {
     }
   }
 
+  private static boolean checkBucketExists(String bucketName) {
+    try {
+      final AmazonS3 bucketCheckHelper = AmazonS3ClientBuilder.defaultClient();
+      return bucketCheckHelper.doesBucketExist(bucketName);
+    } catch (SdkClientException e) {
+      return false;
```
aagarwal-gtr: I am not sure about how this exception is being caught, is this a network error? maybe we should bubble up the actual exception?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -639,6 +645,33 @@ public String toString() {
...
+  }
+
+  private static class BucketValidator implements ConfigDef.Validator {
+    @Override
+    public void ensureValid(String name, Object bucket) {
+      String bucketName = ((String) bucket).toLowerCase().trim();
+      try {
+        BucketNameUtils.validateBucketName(bucketName);
+      } catch (IllegalArgumentException e) {
+        throw new ConnectException(bucketName
```
levzem:
```suggestion
        throw new ConfigException(bucketName
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -639,6 +645,33 @@ public String toString() {
...
+                + " is an invalid bucket name. Does not follow AWS guidelines.");
+      }
+
+      // Checking whether the bucket exists
+      if (!checkBucketExists(bucketName)) {
+        throw new ConnectException("Bucket does not exist.");
```
levzem:
```suggestion
        throw new ConfigException("Bucket " + bucketName + " does not exist.");
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -639,6 +645,33 @@ public String toString() {
...
+      return bucketCheckHelper.doesBucketExist(bucketName);
+    } catch (SdkClientException e) {
+      return false;
+    }
+  }
+
+  private static class BucketValidator implements ConfigDef.Validator {
+    @Override
+    public void ensureValid(String name, Object bucket) {
+      String bucketName = ((String) bucket).toLowerCase().trim();
```
levzem: does case matter for S3 bucket names?

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorConfigTest.java
```diff
@@ -329,5 +340,25 @@ public void testS3RetryBackoffNegative() {
...
+  @Test(expected = ConnectException.class)
+  public void testInvalidBucketName() {
+    properties.put(S3SinkConnectorConfig.S3_BUCKET_CONFIG, "test_bucket");
+    connectorConfig = new S3SinkConnectorConfig(properties);
+  }
+
+  @Test
+  public void testBucketWithValidNameWhichExists() {
+    properties.put(S3SinkConnectorConfig.S3_BUCKET_CONFIG, "test-bucket");
+    connectorConfig = PowerMockito.spy(new S3SinkConnectorConfig(properties));
```
levzem: no need to spy

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorConfigTest.java
```diff
@@ -329,5 +340,25 @@ public void testS3RetryBackoffNegative() {
...
+  }
+
+  @Test(expected = ConnectException.class)
+  public void testBucketWithValidNameAndDoesNotExists() throws Exception {
+    properties.put(S3SinkConnectorConfig.S3_BUCKET_CONFIG, "test-bucket");
+    PowerMockito.doReturn(false).when(S3SinkConnectorConfig.class, "checkBucketExists",
```
levzem: static import
```suggestion
    doReturn(false).when(S3SinkConnectorConfig.class, "checkBucketExists",
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
rhauch: How does the `AmazonS3ClientBuilder` know how to get the credentials? Remember, the `` property might specify a different credential provider class(es), which the `AmazonS3ClientBuilder` may not know about. Or, the `AmazonS3ClientBuilder` might use multiple credential providers but in a different order -- in that case, a user might have locally set them one way (e.g., config file) and still had other credentials set earlier but a different way (e.g., env vars). If the default credentials provider used by the `AmazonS3ClientBuilder` doesn't check them in the same order as the user-defined credentials provider class, then the wrong credentials might be used.

The bottom line is that the Validator on the `s3.bucket.name` only knows the value of that property and simply does not know the value for any other property. And because `s3.credentials.provider.class` is a different property, the `s3.bucket.name` validator can't actually assume the credentials are accessed in any particular way, since that might differ from what the `s3.credentials.provider.class` class says.

PrajwalRavi:
Backported to version 5.0.x. Hence, closing this PR. Sorry for the inconvenience. Please refer to the new PR: [https://github.com/confluentinc/kafka-connect-storage-cloud/pull/304](url)

PrajwalRavi:
Reopening after backporting to 5.0.x.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
```diff
@@ -86,4 +92,36 @@ public ConfigDef config() {
...
+  @Override
+  public Config validate(final Map<String, String> connectorConfigs) {
+    ConfigDef configDef = this.config();
+    if (null == configDef) {
+      throw new ConnectException(String.format("%s.config() must return a ConfigDef that is not"
+              + " null.", this.getClass().getName()));
+    } else {
+
+      // Checking whether the bucket exists
+      if (!checkBucketExists(new S3SinkConnectorConfig(connectorConfigs))) {
```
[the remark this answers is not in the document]
gmg13: also if we do decide to have this, then refactor so that both connector and task uses the same method to create connection and verify for the existence of bucket

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorTest.java
```diff
@@ -38,5 +64,21 @@ public void connectorType() {
     Connector connector = new S3SinkConnector();
     assertTrue(SinkConnector.class.isAssignableFrom(connector.getClass()));
   }
+
+  @Test(expected = ConnectException.class)
+  public void checkNonExistentBucket() throws Exception {
+    configProps.put(S3SinkConnectorConfig.S3_BUCKET_CONFIG, "non-existent-bucket");
+    connector = spy(new S3SinkConnector(new S3SinkConnectorConfig(configProps)));
+    doReturn(false).when(connector, "checkBucketExists", Mockito.any());
```
gmg13: See if it is possible to avoid mocking a private method like so. I think if this is moved to a common lib method, then there will be no need to mock private method. Same applies for the other reverse function. You could then use EasyMock.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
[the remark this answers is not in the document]
PrajwalRavi: validate() is called before the task is created; so we are trying to avoid creating a task if the bucket does not exist. Hence, we need it in the validate() method. 
We are already using a common method defined in the S3Storage class to verify bucket existence. We are just creating an object of S3Storage class using the configs and calling bucketExists() method in the class. The task has this object as a class variable.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
```diff
@@ -86,4 +92,36 @@ public ConfigDef config() {
     return S3SinkConnectorConfig.getConfig();
   }
 
+  private boolean checkBucketExists(S3SinkConnectorConfig config) {
```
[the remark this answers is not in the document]
PrajwalRavi: Moved it to a Util class.

kafka-connect-s3/pom.xml
```diff
@@ -141,7 +141,7 @@
                             <goal>testCompile</goal>
                         </goals>
                         <configuration>
-                            <source>1.7</source>
+                            <source>1.8</source>
```
PrajwalRavi: Was getting ```bootstrap class path not set in conjunction with -source 1.7``` when compiling even though JAVA_HOME and jenv were set to version 1.8. Hence had to make this change.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
```diff
@@ -86,4 +89,15 @@ public ConfigDef config() {
     return S3SinkConnectorConfig.getConfig();
   }
 
+  @Override
+  public Config validate(final Map<String, String> connectorConfigs) {
+    Config config = super.validate(connectorConfigs);
+    // Checking whether the bucket exists
+    if (!S3BucketCheck.checkBucketExists(new S3SinkConnectorConfig(connectorConfigs))) {
+      throw new ConnectException("Bucket does not exist or invalid credentials "
+              + "provided");
```
rhauch: We don't want to throw an exception, because doing so essentially means the validation check could not do its thing. But what we want is for a validation messages to be captured on the bucket property in the `Config` response.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorTest.java
```diff
@@ -38,5 +66,23 @@ public void connectorType() {
     Connector connector = new S3SinkConnector();
     assertTrue(SinkConnector.class.isAssignableFrom(connector.getClass()));
   }
+
+  @Test(expected = ConnectException.class)
```
rhauch: We don't want to get a ConnectException here. Instead, the `Config` returned from the `validate(...)` method should contain our message on the `s3.bucket.name` property.

What happens if the credentials are not correct? Do I get a "bucket does not exist" validation error? If so, should the message should be something like "The S3 bucket named '{}' could not be found. Check the S3 credentials and verify the '{}' bucket exists."

PrajwalRavi: Exactly. If the credentials are incorrect, we get a "bucket does not exist" error. So added the message you suggested.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
```diff
@@ -86,4 +89,21 @@ public ConfigDef config() {
...
+  public Config validate(final Map<String, String> connectorConfigs) {
+    Config config = super.validate(connectorConfigs);
+    // Checking whether the bucket exists
+    if (!S3BucketCheck.checkBucketExists(new S3SinkConnectorConfig(connectorConfigs))) {
+      for (ConfigValue configValue : config.configValues()) {
+        if (configValue.name().equals(S3SinkConnectorConfig.S3_BUCKET_CONFIG)) {
+          String bucketName = configValue.value().toString();
+          configValue.addErrorMessage(String.format("The S3 bucket named %s could not be found."
+                  + " Check the S3 credentials and verify the %s bucket exists.",
+              bucketName, bucketName));
```
rhauch: Nit: our style convention is to separate multi-parameter calls
```suggestion
          configValue.addErrorMessage(
              String.format(
                  "The S3 bucket named %s could not be found."
                  + " Check the S3 credentials and verify the %s bucket exists.",
                  bucketName,
                  bucketName
              )
          );
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -613,6 +616,19 @@ public String toString() {
...
 
+  private static class BucketNameValidator implements ConfigDef.Validator {
+    @Override
+    public void ensureValid(String name, Object bucket) {
+      String bucketName = ((String) bucket).trim();
+      try {
+        BucketNameUtils.validateBucketName(bucketName);
+      } catch (IllegalArgumentException e) {
+        throw new ConfigException(e.getMessage() + ". " + bucketName
+                + " is an invalid bucket name. Does not follow AWS guidelines.");
```
rhauch: Might be better to swap the order and put the most important part first:
```suggestion
        throw new ConfigException(
            String.format(
                "'%s' is not a valid bucket name and does not follow AWS guidelines. %s",
                bucketName,
                e.getMessage()
            )
        );
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -613,6 +616,19 @@ public String toString() {
...
+      }
+    }
```
rhauch: We should always have a `toString()` method on all validators that's used in our documentation, and the string should describe the limitation. For this, something like:
> Follows the AWS guidelines for S3 bucket names

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -613,6 +616,19 @@ public String toString() {
     }
   }
 
+  private static class BucketNameValidator implements ConfigDef.Validator {
```
rhauch: What do you think about moving this into the `S3BucketCheck` class, either as:
```
    public static ConfigDef.Validator bucketNameValidator() {
      return new ConfigDef.Validator() {
        ...
      };
    }
```
or moving the class into `S3BucketCheck` (still as a private static nested class) so that:
```
    public static ConfigDef.Validator bucketNameValidator() {
      return new BucketNameValidator();
    }
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnector.java
```diff
@@ -86,4 +89,25 @@ public ConfigDef config() {
...
+    if (!S3BucketCheck.checkBucketExists(new S3SinkConnectorConfig(connectorConfigs))) {
+      for (ConfigValue configValue : config.configValues()) {
+        if (configValue.name().equals(S3SinkConnectorConfig.S3_BUCKET_CONFIG)) {
+          String bucketName = configValue.value().toString();
+          configValue.addErrorMessage(
+              String.format("The S3 bucket named %s could not be found."
+                      + " Check the S3 credentials and verify the %s bucket exists.",
+                  bucketName,
+                  bucketName
+              )
```
rhauch: Nit: this formatting is still not right. I also suggest wrapping the bucket name with single quotes so that bucket names stand out a bit more in the log message.
```suggestion
              String.format(
                  "The S3 bucket named '%s' could not be found."
                  + " Check the S3 credentials and verify the '%s' bucket exists.",
                  bucketName,
                  bucketName
              )
```

PR #308, opened 2020-03-02
CC-8085: Allow setting compression level for gzip in Json and ByteArray formats

kkonstantine:
This change will allow users to specify a specific level of compression when using the gzip compression type with Json or ByteArray formats.


kkonstantine:
Adding a few test cases shortly

kkonstantine:
Thanks for the review @ncliang !

PR #230, opened 2019-03-04
Add AWS IAM Assume Role credentials provider

SgtPepperLHCB:
Add an AWSCredentialsProvider that uses STSAssumeRoleSessionCredentialsProvider to assume an IAM Role. This allows the worker to write S3 objects that are owned by the bucket owner and not the Kafka Connect worker AWS Account.

PR #311, opened 2020-03-08
Add AWS IAM Assume Role credentials provider

kkonstantine:
Add an AWSCredentialsProvider that uses STSAssumeRoleSessionCredentialsProvider to assume an IAM Role. This allows the worker to write S3 objects that are owned by the bucket owner and not the Kafka Connect worker AWS Account.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
@@ -1,18 +1,17 @@
 /*
- * Copyright 2020 Confluent Inc., Nordstrom Inc.
+ * Copyright 2020 Confluent Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *
  */
 
 package io.confluent.connect.s3.auth;
@@ -73,7 +72,8 @@ public class AwsAssumeRoleCredentialsProvider implements AWSCredentialsProvider,
     return new STSAssumeRoleSessionCredentialsProvider.Builder(roleArn, roleSessionName)
         .withStsClient(AWSSecurityTokenServiceClientBuilder.defaultClient())
         .withExternalId(roleExternalId)
-        .build().getCredentials();
+        .build()
+        .getCredentials();
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


kkonstantine:
This PR is essentially: https://github.com/confluentinc/kafka-connect-storage-cloud/pull/230
with one additional commit to fix minor review comments and a rebase on top of `5.0.x` instead of `5.1.x`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2020 Confluent Inc., Nordstrom Inc.
```
wicknicks: nit: do we mention multiple orgs in the license header?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
```diff
@@ -0,0 +1,84 @@
...
+    roleExternalId = config.getString(ROLE_EXTERNAL_ID_CONFIG);
+    roleSessionName = config.getString(ROLE_SESSION_NAME_CONFIG);
+  }
+
+  @Override
+  public AWSCredentials getCredentials() {
+    return new STSAssumeRoleSessionCredentialsProvider.Builder(roleArn, roleSessionName)
+        .withStsClient(AWSSecurityTokenServiceClientBuilder.defaultClient())
+        .withExternalId(roleExternalId)
+        .build().getCredentials();
```
wicknicks: nit: should `getCredentials()` be on the next line?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
kkonstantine: We do not currently. That was missed. Thanks

kkonstantine:
Good idea @ncliang. Initially I thought those tests it didn't exactly belong in that test class. But works out well. 

Thanks everyone!

sharmargilt:
Hello @kkonstantine , when will this change be available under https://hub.docker.com/r/confluentinc/cp-kafka-connect ?

kkonstantine:
@sharmargilt with any future Confluent Platform release from `5.0.x` or newer (that has included this commit at the time it was being prepared).

moyuanhuang:
Thanks everyone for doing this! We're experiencing the same problem and would love to see this available on docker-hub!

One question: Why this would be released on `5.0.x` instead of later versions like `5.3.x`?

kkonstantine:
Glad to see you're interested @moyuanhuang !

A little bit about how backporting worked here. 
This PR initially targeted, 5.1.x. Since it'll be a useful option, that is self-contained and does not affect other parts of the connector, we made the call to merge it to `5.0.x` to make it available to users that are still installing bugfixes in previous `5.y.x` releases.

What this actually means is that this provider will be available in any of the forthcoming: 
`5.0.x`, `5.1.x`, `5.2.x`, `5.3.x`, `5.4.x` releases and finally the upcoming minor release which will be `5.5.0`. Of course as you noticed, more recent branches get more frequent bugfix releases. Hopefully, one of the next ones will be suitable for you. Of course, I'm saying that with a note that we keep encouraging people to upgrade  to the latest major or minor releases by porting new features in `master`. 

Cheers

moyuanhuang:
@kkonstantine Thanks for the explanation! I think for our use case we'll just use the `latest` tag. ❤️

jssnipes:
@kkonstantine - per 5.3.3 release notes (https://docs.confluent.io/5.3.3/release-notes/index.html#s3-connector) it shows that this PR has made the 5.3.3 release, but I don't see any updated corresponding updates to our documentation.  Can you please confirm?  I have a client on 5.3.1 wanting to use this and would be great if they can upgrade to 5.3.3 to have this functionality.  Thx.

PR #317, opened 2020-03-23
CC-7901: dynamic buffer sizes for output streams

levzem:
This PR adds dynamic buffer size allocation for output streams. It keeps track of the exponentially weighted moving average of the part sizes uploaded and uses that to guess the appropriate buffer size that an output stream will use. If the the buffer is too small, the buffer will be copied into a larger one. 

The measured performance impact was 3% decrease in overall throughput and the connector started at 75% throughput and it took 5 minutes to reach its peak throughput. 

Signed-off-by: Lev Zemlyanov <lev@confluent.io>


hsp8712:
It looks like the same purpose with my PR:https://github.com/confluentinc/kafka-connect-storage-cloud/pull/320. 
But I have not found buffer size shrink when s3 part uploaded.

levzem:
closing for now

PR #318, opened 2020-03-24
CC-8170: Exclude jackson-mapper-asl and commons-beanutils from hadoop-common import

ncliang:
This PR explicitly excludes `jackson-mapper-asl` from being pulled in with `hadoop-common`. Code inspection indicates that the jar is not being used and can be removed. Passing muckrake system tests - https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/3626/


ncliang:
System tests passing with both dependencies removed: https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/3633/

PR #325, opened 2020-04-27
CC-7933: Support writing key and headers to S3 in S3 sink connector

dosvath:
## Summary
This PR adds the ability for the connector to write record keys and headers to separate key and header files if desired. 

The updated code exposes new connector configurations that allow the user to specify if they want to write the contents from the Kafka record key and header to files at the destination in addition to the original value file. 

The connector achieves this by introducing new configs along with a `RecordView` interface and `KeyValueRecordProvider` class.  We introduce our own `RecordWriterProvider` that provides an abstraction so that the `TopicPartitionWriter`  class can write  these new files without any code changes.

## New Classes
**KeyValueHeaderRecordWriterProvider**
This class provides an abstraction for writing, committing and closing all three header, key and value files.

**RecordView**
An interface to get the schema for all three record portions. 

**RecordViews**
Implementation of `RecordView`. 

**RecordViewSetter**
A class to provide functionality to set the recordView for a particular writerProvider, used in `SinkTask::newRecordWriterProvider`.

**Utils**
A general utility class for util functions.

## Functional Changes 
**S3SinkConnectorConfig**
Added the new configurations:
`store.kafka.keys`
`store.kafka.headers`
`keys.format.class` 
`headers.format.class`
and their corresponding recommenders.

**S3SinkTask**
Added the utility function `newRecordWriterProvider` to get the `KeyValueHeaderRecordWriterProvider` with the given headerWriterProvider and keyWriterProviders if storing the headers/keys is configured. 

**ParquetRecordWriterProvider**
Extend `RecordViewSetter` to use the default record view as value. Added filename manipulation logic to append the key/header extension if needed.

**JsonRecordWriterProvider**
Extend `RecordViewSetter` to use the default record view as value. Added filename manipulation logic to append the key/header extension if needed.

**ByteArrayRecordWriterProvider**
Extend `RecordViewSetter` to use the default record view as value. Added filename manipulation logic to append the key/header extension if needed.

**AvroRecordWriterProvider**
Extend `RecordViewSetter` to use the default record view as value. Added filename manipulation logic to append the key/header extension if needed.

## Tests
**Muckrake Tests** https://github.com/confluentinc/muckrake/pull/1020 (System tests PR)

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2017 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.confluent.connect.s3.format;
+
+import org.apache.kafka.connect.sink.SinkRecord;
+
+import io.confluent.connect.s3.S3SinkConnectorConfig;
+import io.confluent.connect.storage.format.RecordWriter;
+import io.confluent.connect.storage.format.RecordWriterProvider;
+
+import static java.util.Objects.requireNonNull;
+
+public class KeyValueHeaderRecordWriterProvider
+    implements RecordWriterProvider<S3SinkConnectorConfig> {
+
+  private final RecordWriterProvider<S3SinkConnectorConfig> valueProvider;
+  private final RecordWriterProvider<S3SinkConnectorConfig> keyProvider;
+  private final RecordWriterProvider<S3SinkConnectorConfig> headerProvider;
+
+
+  public KeyValueHeaderRecordWriterProvider(
+      RecordWriterProvider<S3SinkConnectorConfig> valueProvider,
+      RecordWriterProvider<S3SinkConnectorConfig> keyProvider,
+      RecordWriterProvider<S3SinkConnectorConfig> headerProvider) {
+    this.valueProvider = requireNonNull(valueProvider);
+    this.keyProvider = keyProvider;
+    this.headerProvider = headerProvider;
+  }
+
+  @Override
+  public String getExtension() {
+    return valueProvider.getExtension();
+  }
+
+  @Override
+  public RecordWriter getRecordWriter(S3SinkConnectorConfig conf, String filename) {
+    RecordWriter valueWriter = valueProvider.getRecordWriter(conf, filename);
+    RecordWriter keyWriter =
+        keyProvider == null ? null : keyProvider.getRecordWriter(conf, filename);
+    RecordWriter headerWriter =
+        headerProvider == null ? null : headerProvider.getRecordWriter(conf, filename);
+
+    return new RecordWriter() {
+      @Override
+      public void write(SinkRecord sinkRecord) {
+        valueWriter.write(sinkRecord);
+        if (keyWriter != null) {
+          keyWriter.write(sinkRecord);
+        }
+        if (headerWriter != null) {
+          headerWriter.write(sinkRecord);
+        }
+      }
+
+      @Override
+      public void close() {
+        valueWriter.close();
+        if (keyWriter != null) {
+          keyWriter.close();
+        }
+        if (headerWriter != null) {
+          headerWriter.close();
+        }
+      }
+
+      @Override
+      public void commit() {
+        valueWriter.commit();
+        if (keyWriter != null) {
+          keyWriter.commit();
+        }
+        if (headerWriter != null) {
+          headerWriter.commit();
+        }
+      }
+    };
+  }
+}
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
@@ -67,8 +67,9 @@ public class KeyValueHeaderRecordWriterProvider
   @Override
   public RecordWriter getRecordWriter(S3SinkConnectorConfig conf, String filename) {
     // Remove extension to allow different formats for value, key and headers.
-    // Each provider will add its own extension. The filename comes in with the value file format, e.g. filename.avro,
-    // but when the format class is different for the key or the headers the extension needs to be removed.
+    // Each provider will add its own extension. The filename comes in with the value file format,
+    // e.g. filename.avro, but when the format class is different for the key or the headers the
+    // extension needs to be removed.
     int extensionIndex = filename.indexOf(valueProvider.getExtension());
     String strippedFilename = extensionIndex > -1
         ? filename.substring(0, extensionIndex)
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
@@ -17,6 +17,8 @@
 package io.confluent.connect.s3.format;
 
 
+import javax.annotation.Nullable;
+import javax.validation.constraints.NotNull;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.sink.SinkRecord;
 
@@ -28,29 +30,30 @@ import org.slf4j.LoggerFactory;
 
 import static java.util.Objects.requireNonNull;
 
+/**
+ * A class that adds a record writer layer to manage writing values, keys and headers
+ * with a single call. It provides an abstraction for writing, committing and
+ * closing all three header, key and value files.
+ */
 public class KeyValueHeaderRecordWriterProvider
     implements RecordWriterProvider<S3SinkConnectorConfig> {
 
   private static final Logger log =
       LoggerFactory.getLogger(KeyValueHeaderRecordWriterProvider.class);
 
-  /**
-   * valueProvider may not be null.
-   */
+  @NotNull
   private final RecordWriterProvider<S3SinkConnectorConfig> valueProvider;
-  /**
-   * keyProvider may be null.
-   */
+
+  @Nullable
   private final RecordWriterProvider<S3SinkConnectorConfig> keyProvider;
-  /**
-   * headerProvider may be null.
-   */
+
+  @Nullable
   private final RecordWriterProvider<S3SinkConnectorConfig> headerProvider;
 
   public KeyValueHeaderRecordWriterProvider(
       RecordWriterProvider<S3SinkConnectorConfig> valueProvider,
-      RecordWriterProvider<S3SinkConnectorConfig> keyProvider,
-      RecordWriterProvider<S3SinkConnectorConfig> headerProvider) {
+      @Nullable RecordWriterProvider<S3SinkConnectorConfig> keyProvider,
+      @Nullable RecordWriterProvider<S3SinkConnectorConfig> headerProvider) {
     this.valueProvider = requireNonNull(valueProvider);
     this.keyProvider = keyProvider;
     this.headerProvider = headerProvider;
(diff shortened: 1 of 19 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
```diff
@@ -56,15 +61,24 @@ public String getExtension() {
...
       @Override
       public void write(SinkRecord record) {
-        log.trace("Sink record: {}", record);
+        log.trace("Sink record with view {}: {}", recordView, record);
         try {
-          byte[] bytes = converter.fromConnectData(
-              record.topic(), record.valueSchema(), record.value());
+          byte[] bytes;
+          if (recordView instanceof HeaderRecordView) {
+            JsonConverter jsonConverter = new JsonConverter();
```
kkonstantine, line 74: we can't create a new converter instance with every record. 
We have to cache it.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
```diff
@@ -57,19 +67,29 @@ public String getExtension() {
...
       public void write(SinkRecord record) {
         if (schema == null) {
-          schema = record.valueSchema();
+          schema = recordView.getViewSchema(record);
+          // pad the schema for AvroParquetWriter that only accepts record types
+          if (recordView instanceof HeaderRecordView) {
+            schema = SchemaBuilder.struct().name(HEADER_STRUCT_NAME)
+                .field(HEADER_FIELD_NAME, schema).build();
+          } else if (recordView instanceof KeyRecordView) {
+            schema = SchemaBuilder.struct().name(KEY_STRUCT_NAME)
```
kkonstantine, line 84: how do we know that they key at least is not of record type? If someone chooses parquet for keys, we'd assume they have they keys in the appropriate type (like the values).

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
```diff
@@ -82,8 +102,15 @@ public void write(SinkRecord record) {
...
         }
-        log.trace("Sink record: {}", record.toString());
-        Object value = avroData.fromConnectData(schema, record.value());
+        log.trace("Sink record with view {}: {}", recordView, record);
+        Object view = recordView.getView(record);
+        // fill the schema if it was padded
+        if (recordView instanceof HeaderRecordView) {
+          view = new Struct(schema).put(HEADER_FIELD_NAME, view);
+        } else if (recordView instanceof KeyRecordView) {
+          view = new Struct(schema).put(KEY_FIELD_NAME, view);
```
kkonstantine, line 111: same question as above for keys

kkonstantine:
Also, please rebase on top of the most recent changes.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
dosvath: updated code with a check for the type

dosvath:
@kkonstantine Addressed comments and rebased. (commits appear on Github PR twice because of the rebase)

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
```diff
@@ -56,15 +63,23 @@ public String getExtension() {
...
       @Override
       public void write(SinkRecord record) {
-        log.trace("Sink record: {}", record);
+        log.trace("Sink record with view {}: {}", recordView, record);
         try {
-          byte[] bytes = converter.fromConnectData(
-              record.topic(), record.valueSchema(), record.value());
+          byte[] bytes;
+          if (recordView instanceof HeaderRecordView) {
+            jsonConverter.configure(Collections.singletonMap("schemas.enable", false), false);
```
ncliang, line 76: This special handling for trying to encode headers as `byte[]` is awkward. I don't know if people would actually try to do that. Could we just say you can't use `ByteArrayFormat` for encoding headers? Remove it from the config recommender, etc?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
```diff
@@ -57,19 +68,33 @@ public String getExtension() {
...
       ParquetWriter<GenericRecord> writer;
       S3ParquetOutputFile s3ParquetOutputFile;
 
       @Override
       public void write(SinkRecord record) {
+        boolean schemaPadded = false;
         if (schema == null) {
-          schema = record.valueSchema();
+          schema = recordView.getViewSchema(record);
+          // pad the schema for AvroParquetWriter that only accepts record types
```
ncliang, line 81: This special handling here is also awkward. Is it possible to do these in the corresponding `RecordView` implementations instead?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
dosvath: If we use `ByteArrayFormat` for the `headers.format.class` this is still the record provider that gets called. Did you mean to say `JsonFormat`? @kkonstantine may have some input on how we'd like to handle this.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
dosvath: Do you mean the padding logic?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
ncliang: Well, this code is for handling if `ByteArrayFormat` is specified in `headers.format.class`. I don't know if that is actually useful. If not, we could just not support it and remove this one-off special case handling code.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
ncliang: Yeah. Doing this one-off handling in parquet record writer is awkward since it has nothing to do with writing parquet records per-say. Instead, it is related to what "view" of the record we want the parquet record writer to see and write. I think this could be done in the `RecordView` by modifying the interface to allow for specifying if the view should be padded or not. i.e.
```
public interface RecordView {
  Schema getViewSchema(SinkRecord record, boolean enveloped);
  Object getView(SinkRecord record, boolean enveloped);
}
```
Also, just a nit: I think of this sort of wrapper object as an envelope. Padding I think of extra empty bytes in a byte[].

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
dosvath: I see what you mean now. Yeah that may be an option. @kkonstantine

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
dosvath: Yes that can work, for some we may not need to use the `boolean` at all because we don't want to envelop, but that's not an issue. I'll update the PR.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
kkonstantine: I agree, let's remove support for headers as a byte array for now, since the use of json converter and encoding is a bit arbitrary.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/bytearray/ByteArrayRecordWriterProvider.java
```diff
@@ -46,6 +52,7 @@
     this.lineSeparatorBytes = storage.conf()
         .getFormatByteArrayLineSeparator()
         .getBytes(StandardCharsets.UTF_8);
+    this.jsonConverter = new JsonConverter();
```
ncliang, line 55: remove

dosvath:
Thanks @ncliang, updated. @kkonstantine should we target `master` instead of `5.5.x` because this is a new feature? It would align with the DLQ work targeting `master`.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
```diff
@@ -0,0 +1,131 @@
...
+    return valueProvider.getExtension();
+  }
+
+  @Override
+  public RecordWriter getRecordWriter(S3SinkConnectorConfig conf, String filename) {
+    // Remove extension to allow different formats for value, key and headers
+    // each provider will add its own extension.
+    // The filename comes in with the value file format, ex. a.avro,
+    // but when the format class is different for the keys,
+    // header this needs to be removed.
```
kkonstantine, line 73:
```suggestion
    // Remove extension to allow different formats for value, key and headers.
    // Each provider will add its own extension. The filename comes in with the value file format, e.g. filename.avro,
    // but when the format class is different for the key or the headers the extension needs to be removed.
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
```diff
@@ -0,0 +1,131 @@
...
+    RecordWriter headerWriter =
+        headerProvider == null ? null : headerProvider.getRecordWriter(conf, strippedFilename);
+
+    return new RecordWriter() {
+      @Override
+      public void write(SinkRecord sinkRecord) {
+        valueWriter.write(sinkRecord); // null check happens in sink task
+        // keyWriter != null means writing keys is turned on
+        if (keyWriter != null && sinkRecord.key() == null) {
+          throw new ConnectException(
```
kkonstantine, line 91: should probably be `DataException`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
```diff
@@ -0,0 +1,131 @@
...
+              String.format("Key cannot be null for SinkRecord: %s", sinkRecord)
+          );
+        } else if (keyWriter != null) {
+          keyWriter.write(sinkRecord);
+        }
+
+        // headerWriter != null means writing headers is turned on
+        if (headerWriter != null && sinkRecord.headers() == null) {
+          throw new ConnectException(
```
kkonstantine, line 100: `DataException` as above

kkonstantine (commented):
Almost there! 
A suggestion to make a comment easier to understand and using `DataException`s on `null` values for keys and headers.

dosvath:
@kkonstantine thanks for the review, updated. I will do a final system test run on this branch before merging to double check. It's looking ready for approval if there's nothing else 👍

kkonstantine (approved):
Thanks @dosvath 
This LGTM

oripwk:
Just a quick question - does it also have the ability to add the headers as a nested object to the file where the value itself is written?

I don't see much use if it's in a separate file. For example in our use case we read the resulted Parquet files with Presto and there's little value if the headers are in separate files since they logically belong to the value's record.

dosvath:
> Just a quick question - does it also have the ability to add the headers as a nested object to the file where the value itself is written?
> 
> I don't see much use if it's in a separate file. For example in our use case we read the resulted Parquet files with Presto and there's little value if the headers are in separate files since they logically belong to the value's record.

Hi @oripwk the functionality you mention of including the headers in the value file is not part of this PR. I will check with my team on the potential of that feature.

kkonstantine:
@oripwk making general enough assumptions that can cover a wide range of use cases w/r/t the format of your keys, headers and values and doing so in a backwards compatible way is not straightforward. 

One easier avenue would be for each user to define their own envelop format (schema) and add keys, headers and values in the record value. This can be achieved as an SMT and be used with the official connector, without any code changes being required.

Issue #247, opened 2019-05-14
Support multiple schemas in a topic


papa99do:
Kafka supports multiple types of schema in one topic with this PR: https://github.com/confluentinc/schema-registry/pull/680. This introduced a problem for S3 connector, since every time it encounters a different type of schema, it will create a new file in S3 bucket, which will result in tons of small files in this case.

Can we support this by writing different types of events in the same topic to separate S3 files ( with the schemas full qualified name in the path of the file)?

papa99do:
Just to check if there's any update on this ticket?
`io.confluent.connect.s3.TopicPartitionWriter` will rotate a file whenever the compatibility check (with the last schema) fails. And the name of a file only has `tp.topic()` in it. In order to support mutliple types of schema in one topic, we might need to create a new `TopicPartitionWriter` which is message-type-aware.

Any thoughts? Or any suggestions how I can work around this. It looks to me a general problem for anyone having multiple types in a topic.

teabot:
As a workaround, how feasible would it be to set up one connector per type and apply a type specific filter in each? Obviously not great from a resource usage perspective as it'll require `N` consumers for `N` types, but I'd be keen to know the feasibility of this approach.

corbinbs:
One solution for getting the schema name in the S3 key is to create a custom storage partitioner class.  For example, if you extend `io.confluent.connect.storage.partitioner.TimeBasedPartitioner` and override the `encodePartition` methods, you can get the schema name via `sinkRecord.valueSchema().name()`.  
```
sinkRecord.valueSchema().name() + this.delim + timeBasedPartitionerEncodedPartition
```
that would result in something like `topics/{topic_name}/{schema_name}/{time_partition}/file.avro` in the bucket that the S3 Sink is targeting.

Of course, this doesn't address the many small files that could be created if the topic has a lot of events alternating back and forth between different types.  If you can tolerate that, this approach might help get multiple event types from a single topic organized in the S3 bucket so that it'd be suitable for querying with something like AWS Athena, etc.  Once the files land in S3, you could periodically roll them up into more optimized file sizes to help control the number of files.

papa99do:
Hi @corbinbs , Thanks for the suggestion. That's a brilliant idea. The only issue was the fragmented files which I managed to work around using the following changes. 

[...]

PR #328, opened 2020-05-13
Move null value handling earlier in SinkTask put

mageshn:
Why
-----
TopicPartitionWriter threw "SchemaProjectorException( "Switch between schema-based and schema-less data is not supported");" from storage common, despite having an AvroConverter, which we would expect to require every record to have a schema. The exceptional case, it turns out, is a null-valued schema which can have no schema. This alerted us that our original implementation of null-value-behavior in the writerprovider was carrying an assumption that the null-valued record would be well-handled by the topicpartitionwriter, which in this case, is not.

How
------
Moved the `nullValueBehavior` logic from the WriterProvider to be much earlier in Task.put, so we can continue to the next record before it can throw escalations or waste resources.

Testing
-----
Unit test included for regressions.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -177,12 +177,6 @@ public class S3SinkTask extends SinkTask {
       TopicPartition tp = new TopicPartition(topic, partition);
 
       if (maybeSkipOnNullValue(record)) {
-        log.debug(
-            "Null valued record from topic '{}', partition {} and offset {} was skipped.",
-            record.topic(),
-            record.kafkaPartition(),
-            record.kafkaOffset()
-        );
         continue;
       }
       topicPartitionWriters.get(tp).buffer(record);
@@ -213,6 +207,12 @@ public class S3SinkTask extends SinkTask {
     if (record.value() == null) {
       if (connectorConfig.nullValueBehavior()
           .equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
+        log.debug(
+            "Null valued record from topic '{}', partition {} and offset {} was skipped.",
+            record.topic(),
+            record.kafkaPartition(),
+            record.kafkaOffset()
+        );
         return true;
       } else {
         throw new ConnectException("Null valued records are not writeable with current "
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -176,7 +176,8 @@ public class S3SinkTask extends SinkTask {
       int partition = record.kafkaPartition();
       TopicPartition tp = new TopicPartition(topic, partition);
       if (record.value() == null) {
-        if (connectorConfig.nullValueBehavior().equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
+        if (connectorConfig.nullValueBehavior()
+            .equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
           log.debug("Null valued record cannot be written to output as Avro. "
               + "Skipping. Record Key: {}", record.key());
           return;
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -178,8 +178,8 @@ public class S3SinkTask extends SinkTask {
       if (record.value() == null) {
         if (connectorConfig.nullValueBehavior()
             .equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
-          log.debug("Null valued record cannot be written to output as Avro. "
-              + "Skipping. Record Key: {}", record.key());
+          log.debug("Null valued record cannot be written to output: offset {}",
+              record.kafkaOffset());
           continue;
         } else {
           throw new ConnectException("Null valued records are not writeable with current "
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -177,9 +177,12 @@ public class S3SinkTask extends SinkTask {
       TopicPartition tp = new TopicPartition(topic, partition);
 
       if (maybeSkipOnNullValue(record)) {
-        log.debug("Null valued record cannot be written to output: offset {}",
-            record.kafkaOffset());
-        continue;
+        log.debug(
+            "Null valued record from topic '{}', partition {} and offset {} was skipped.",
+            record.topic(),
+            record.kafkaPartition(),
+            record.kafkaOffset()
+        );
       }
       topicPartitionWriters.get(tp).buffer(record);
     }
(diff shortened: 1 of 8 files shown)


cyrusv:
We'll need to do this for all storage connectors

cyrusv:
Also we need a regression test -- @wicknicks can you share your reproduction code?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -183,6 +184,17 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
       String topic = record.topic();
       int partition = record.kafkaPartition();
       TopicPartition tp = new TopicPartition(topic, partition);
+      if (record.value() == null) {
```
kkonstantine, line 187: I suggest fleshing out in a `maybeSkipOnNullValue` method that returns a `boolean`

cyrusv:
@kkonstantine we are returning to since TopicPartitionWriter threw  "SchemaProjectorException(          "Switch between schema-based and schema-less data is not supported");" from storage common, despite having an AvroConverter, which we would expect to require every record to have a schema. The exceptional case, it turns out, is a null-valued schema which can have no schema. This alerted us that our original implementation of null-value-behavior in the writerprovider was carrying an assumption that the null-valued record would be well-handled by the topicpartitionwriter, which in this case, is not.

mageshn:
I will be closing this PR. I wanted to make a quick patch for the prod issue.

cyrusv:
@mageshn can you fill out the description fields in the PR the future?

cyrusv:
And mark as "WIP" or "DRAFT" in the

cyrusv:
@kkonstantine @wicknicks  I'm picking up this PR and carrying it to completion -- looking forward to your feedback!

I think @kkonstantine suggestion to remove the logic from the writer-provider is wise. I will make that change

cyrusv:
I have a more fundamental question for @kkonstantine: Why can't we consider schema'd and schema-less records to be false for shouldChangeSchema?

kkonstantine:
@cyrusv I'm not sure I understand the question. Could you bring an example?

Thanks for the description @mageshn . Makes sense now.

cyrusv:
@kkonstantine TopicPartitionWriter calls StorageSchemaCompatibility.shouldChangeSchema, which returns `true` if the schemas are different, `false` if they are the same, and throws this error if one has schema and the other does not have schema. This makes sense to me, but I'm wondering if it's overly strict: Can we consider a schema and schemaless record, can I just return true in this method, and rotate in topicpartitionwriter, and achieve the intention expected by the user?

An example: If I put a bunch of records with the same schema, then a bunch of records with no schema, why not put the ones with the same schema into an output file together, and the schema-less records in a file together? As opposed to what we do today, which I'm thinking may be overly strict: to throw an error

kkonstantine:
@cyrusv I guess the quickest answer is because you can't change formatter on the fly, and most formatters expect a schema. The intention of throwing that exception is that there's no obvious use case for something like this. I can imagine you'd break downstream systems too. 

Why would you have two files in the same partition with and without schemas. 
This https://github.com/confluentinc/kafka-connect-storage-cloud/issues/247 on the other hand seems like a more real use case. 

Is this only to better support tombstone records?

cyrusv:
The most immediate problem is this different type of tombstone. It got me wondering though why the restriction exists at all. It's a good point about changing formatters.

I am not sure I can think of a reason to intentionally do this, but the runtime exception can be quite disruptive from an operational perspective if an unexpected record happens to get into the topic, since we don't have broker-side validation for schemas in a topic. In this case, what would you recommend I do? The record is already in the topic, I'd prefer not to create a new topic. Only other option would be to use a filter SMT.

I do see you're making the point that a downstream system could break, but I don't know if it's the responsibility of the S3 connector to be opinionated about what might or might not be healthy for a downstream system.

I don't feel too strongly about this, so if it complicates the discussion too much we can focus on the schemaless null-valued records that get mixed in with avro records.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -183,6 +184,17 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
...
+        if (connectorConfig.nullValueBehavior()
+            .equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
+          log.debug("Null valued record cannot be written to output as Avro. "
+              + "Skipping. Record Key: {}", record.key());
```
wicknicks, line 191: probably print the topic, partition or offset. sending the key into logs may be unsafe.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkTaskTest.java
```diff
@@ -100,6 +102,30 @@ public void testWriteRecord() throws Exception {
     verify(sinkRecords, validOffsets);
   }
 
+  @Test
+  public void testWriteTwoRecords() throws Exception {
```
wicknicks, line 106: might need to rename this test. I think the point here was to test if the task can survive a null valued record? so maybe `testPutNullRecord`, and maybe we can send maybe 5-6 records after that and check the records that were sent before and after the null were successfully received at the (mock) destination? you can use `createRecords` to generate some sample records.

cyrusv:
Thanks for the feedback, I've made some updates

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
kkonstantine: cc @cyrusv

kkonstantine:
@cyrusv consolidating the method outside `put` still makes sense to me. 
Also, we'd need to resolve conflicts and consider backporting.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -183,6 +184,17 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
       String topic = record.topic();
       int partition = record.kafkaPartition();
       TopicPartition tp = new TopicPartition(topic, partition);
+      if (record.value() == null) {
+        if (connectorConfig.nullValueBehavior()
+            .equalsIgnoreCase(BehaviorOnNullValues.IGNORE.toString())) {
+          log.debug("Null valued record cannot be written to output: offset {}",
+              record.kafkaOffset());
```
wicknicks, line 191: nit, might be useful to print the topic partition also. how about: 

```suggestion
        log.debug(
            "Null valued record from topic '{}', partition {} and offset {} was skipped.",
            record.topic(),
            record.kafkaPartition(),
            record.kafkaOffset()
        );
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
cyrusv: addressed now!

cyrusv:
@kkonstantine I've taken your suggestion to separate to a private method, and I've targetted 5.4.x which was where the null-behavior config was introduced

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -174,6 +175,16 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
       String topic = record.topic();
       int partition = record.kafkaPartition();
       TopicPartition tp = new TopicPartition(topic, partition);
+
+      if (maybeSkipOnNullValue(record)) {
+        log.debug(
```
kkonstantine, line 180: hmm, why did we keep the log message outside the helper method?

PR #340, opened 2020-06-25
CC-9753: Validate timezone configuration with rotate.schedule.interval.ms

dosvath:
## Problem
The `timezone` configuration is not being validated when the `TimeBasedPartitioner` is not set. This results in a runtime error when the `rotate.schedule.interval.ms` is used and an invalid or no `timezone` value is set. 

## Solution
Validate the `timezone` configuration whenever set, and expect the `timezone` configuration to be set when `rotate.schedule.interval.ms` is used. 

- [X] yes
- [ ] no

##### If yes, where?
All storage sink connectors that use the timezone configuration.

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [X] Manual tests

## Release Plan

 
Backporting to TBD.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -530,6 +530,21 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
     addToGlobal(partitionerConfig);
     addToGlobal(commonConfig);
     addToGlobal(this);
+    validateTimezone();
+  }
+
+  private void validateTimezone() {
+    String timezone = getString(PartitionerConfig.TIMEZONE_CONFIG);
+    long rotateScheduleIntervalMs = getLong(ROTATE_SCHEDULE_INTERVAL_MS_CONFIG);
+    if (rotateScheduleIntervalMs > 0 && timezone.isEmpty()) {
+      throw new ConfigException(
+          String.format(
+              "%s configuration must be set when using %s",
+              PartitionerConfig.TIMEZONE_CONFIG,
+              ROTATE_SCHEDULE_INTERVAL_MS_CONFIG
+          )
+      );
+    }
   }
 
   private void addToGlobal(AbstractConfig config) {
--- a/kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorConfigTest.java
+++ b/kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorConfigTest.java
@@ -33,7 +33,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import io.confluent.connect.s3.auth.AwsAssumeRoleCredentialsProvider;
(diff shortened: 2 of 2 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -586,6 +587,34 @@ protected S3SinkConnectorConfig(ConfigDef configDef, Map<String, String> props)
...
+        DateTimeZone.forID(timezone);
+      } catch (IllegalArgumentException e) {
+        throw new ConfigException(
+            String.format("%s configuration invalid: %s",
+                PartitionerConfig.TIMEZONE_CONFIG,
+                e.getMessage()
+            )
+        );
+      }
+    }
```
levzem, line 606: make this an actual validator like `RegionValidator` that validates the actual config

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -586,6 +587,34 @@ protected S3SinkConnectorConfig(ConfigDef configDef, Map<String, String> props)
...
+
+    long rotateScheduleIntervalMs =
+        getLong(S3SinkConnectorConfig.ROTATE_SCHEDULE_INTERVAL_MS_CONFIG);
+    if (rotateScheduleIntervalMs > 0 && timezone.isEmpty()) {
+      throw new ConfigException(
+          String.format("%s configuration must be set when using %s",
```
levzem, line 612: nit
```suggestion
          String.format(
               "%s configuration must be set when using %s",
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -586,6 +587,34 @@ protected S3SinkConnectorConfig(ConfigDef configDef, Map<String, String> props)
...
+            String.format("%s configuration invalid: %s",
+                PartitionerConfig.TIMEZONE_CONFIG,
+                e.getMessage()
+            )
+        );
+      }
+    }
+
+    long rotateScheduleIntervalMs =
+        getLong(S3SinkConnectorConfig.ROTATE_SCHEDULE_INTERVAL_MS_CONFIG);
```
levzem, line 609:
```suggestion
        getLong(ROTATE_SCHEDULE_INTERVAL_MS_CONFIG);
```

dosvath:
> if we do this check in `PartitionerConfig` in `kafka-connect-storage-common` we can have this validator available to all the connectors

That would be great, the only issue is that we need `ROTATE_SCHEDULE_INTERVAL_MS_CONFIG` from the `S3SinkConnectorConfig` for the second check, which is not available in `PartitionerConfig`. An alternative would be to do the Part 1 of the check in `PartitionerConfig` and the second part in the S3 config so we have access to `ROTATE_SCHEDULE_INTERVAL_MS_CONFIG`. At that point we'd still be making changes to each individual connector that has a config similar to the `ROTATE_SCHEDULE_INTERVAL_MS_CONFIG` that needs to be validated in conjunction with the `TIMEZONE`. Both approaches work, depends on how we want to implement these. Having Part 1 of the check (for the timezone validation only) in `PartitionerConfig` would help keep things consistent with the validation across connectors, one concern with that is whether all connectors are expected to validate the timezone config using the `org.joda.time.DateTimeZone` library.

levzem:
I think doing part 1 in `PartitionerConfig` would be good for consistency. I don't think there is an issue with `DateTimeZone` validation but would be good to check

dosvath:
> I think doing part 1 in `PartitionerConfig` would be good for consistency. I don't think there is an issue with `DateTimeZone` validation but would be good to check

Sounds good. [Pushed part 1 ](https://github.com/confluentinc/kafka-connect-storage-common/pull/136) to the `kafka-connect-storage-common` repo to have the fix for all connectors. Updating this PR once we have a version with the parent out.

dosvath:
@levzem Updated the PR with only the secondary check, the `kafka-connect-storage-common` version is usually updated along with the next CP release, so there is no version bump necessary for `kafka-connect-storage-common-parent` in this PR. 

For the validator approach you mentioned, having the timezone check similarly to the `RegionValidator` is not possible here as both `timezone` and `rotate.schedule.interval.ms` configs are defined in different files and repos, unlike the `s3.region` config. From what I've seen, I think we need the `configDef.define` in order to apply a validator, and we don't have that here for any of the two related configs. Let me know if you have any other ideas.

PR #349, opened 2020-08-04
CC-9723: DLQ support in S3 sink

dosvath:
## Problem
The S3 Sink connector currently does not report faulty records to the DLQ. Having this functionality provides more information on failed records and is useful for debugging and monitoring purposes.

## Solution
Add the `ErrantRecordReporter` introduced by [KIP-610](https://cwiki.apache.org/confluence/display/KAFKA/KIP-610%3A+Error+Reporting+in+Sink+Connectors) that writes failed records and their errors to the specified DLQ topic. The topic name can be specified by `errors.deadletterqueue.topic.name`.

### Writes to DLQ in the following cases
1. Record value is `null` and `behavior.on.null.values` is `IGNORE`. (S3SinkTask)
2. The record was determined to be malformed on attempting write. (TopicPartitionWriter)
3. The record key is `null` but `write.record.keys` is turned on.
4. The record headers are `null` but `write.record.headers` is turned on. 

- [x] yes
- [ ] no

##### If yes, where?
In storage sink connectors that don't yet implement this functionality. 

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
`master` for this new feature, and due to `ErrantRecordReporter` dependency.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -20,7 +20,6 @@ import io.confluent.connect.s3.S3SinkConnectorConfig.BehaviorOnNullValues;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.errors.ConnectException;
-import org.apache.kafka.connect.errors.DataException;
 import org.apache.kafka.connect.errors.RetriableException;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.sink.SinkTask;
@@ -108,7 +107,7 @@ public class S3SinkTask extends SinkTask {
           url
       );
       if (!storage.bucketExists()) {
-        throw new DataException("No-existent S3 bucket: " + connectorConfig.getBucketName());
+        throw new ConnectException("No-existent S3 bucket: " + connectorConfig.getBucketName());
       }
 
       writerProvider = newRecordWriterProvider(connectorConfig);
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -20,7 +20,9 @@ import io.confluent.connect.s3.S3SinkConnectorConfig.BehaviorOnNullValues;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.errors.DataException;
 import org.apache.kafka.connect.errors.RetriableException;
+import org.apache.kafka.connect.sink.ErrantRecordReporter;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.sink.SinkTask;
 import org.apache.kafka.connect.sink.SinkTaskContext;
@@ -59,6 +61,7 @@ public class S3SinkTask extends SinkTask {
   private Format<S3SinkConnectorConfig, String> format;
   private RecordWriterProvider<S3SinkConnectorConfig> writerProvider;
   private final Time time;
+  private ErrantRecordReporter reporter;
 
   /**
    * No-arg constructor. Used by Connect framework.
@@ -114,6 +117,16 @@ public class S3SinkTask extends SinkTask {
       partitioner = newPartitioner(connectorConfig);
 
       open(context.assignment());
+      try {
+        reporter = context.errantRecordReporter();
+        if (reporter == null) {
+          log.info("Errant record reporter not configured.");
+        }
+      } catch (NoSuchMethodError | NoClassDefFoundError | UnsupportedOperationException e) {
+        // Will occur in Connect runtimes earlier than 2.6
+        log.warn("AK versions prior to 2.6 do not support the errant record reporter");
+      }
+
       log.info("Started S3 connector task with assigned partitions: {}",
           topicPartitionWriters.keySet()
       );
@@ -203,6 +216,9 @@ public class S3SinkTask extends SinkTask {
       TopicPartition tp = new TopicPartition(topic, partition);
 
       if (maybeSkipOnNullValue(record)) {
+        if (reporter != null) {
+          reporter.report(record, new DataException("Cannot write null value record"));
+        }
         continue;
       }
       topicPartitionWriters.get(tp).buffer(record);
@@ -304,7 +320,8 @@ public class S3SinkTask extends SinkTask {
         partitioner,
         connectorConfig,
         context,
-        time
+        time,
+        reporter
     );
   }
 
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -124,7 +124,8 @@ public class S3SinkTask extends SinkTask {
         }
       } catch (NoSuchMethodError | NoClassDefFoundError | UnsupportedOperationException e) {
         // Will occur in Connect runtimes earlier than 2.6
-        log.warn("AK versions prior to 2.6 do not support the errant record reporter");
+        log.warn("Connect versions prior to Apache Kafka 2.6 do not support "
+            + "the errant record reporter");
       }
 
       log.info("Started S3 connector task with assigned partitions: {}",
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -218,7 +218,7 @@ public class S3SinkTask extends SinkTask {
 
       if (maybeSkipOnNullValue(record)) {
         if (reporter != null) {
-          reporter.report(record, new DataException("Cannot write null value record"));
+          reporter.report(record, new DataException("Skipping null value record."));
         }
         continue;
       }
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -110,7 +110,7 @@ public class S3SinkTask extends SinkTask {
           url
       );
       if (!storage.bucketExists()) {
-        throw new ConnectException("No-existent S3 bucket: " + connectorConfig.getBucketName());
+        throw new ConnectException("Non-existent S3 bucket: " + connectorConfig.getBucketName());
       }
 
       writerProvider = newRecordWriterProvider(connectorConfig);
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -22,6 +22,7 @@ import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.errors.IllegalWorkerStateException;
 import org.apache.kafka.connect.errors.SchemaProjectorException;
+import org.apache.kafka.connect.sink.ErrantRecordReporter;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.kafka.connect.sink.SinkTaskContext;
 import org.joda.time.DateTime;
@@ -91,24 +92,31 @@ public class TopicPartitionWriter {
   private DateTimeZone timeZone;
   private final S3SinkConnectorConfig connectorConfig;
   private static final Time SYSTEM_TIME = new SystemTime();
+  private ErrantRecordReporter reporter;
 
-  public TopicPartitionWriter(TopicPartition tp,
-                              S3Storage storage,
-                              RecordWriterProvider<S3SinkConnectorConfig> writerProvider,
-                              Partitioner<?> partitioner,
-                              S3SinkConnectorConfig connectorConfig,
-                              SinkTaskContext context) {
-    this(tp, storage, writerProvider, partitioner, connectorConfig, context, SYSTEM_TIME);
+  public TopicPartitionWriter(
+      TopicPartition tp,
+      S3Storage storage,
+      RecordWriterProvider<S3SinkConnectorConfig> writerProvider,
+      Partitioner<?> partitioner,
+      S3SinkConnectorConfig connectorConfig,
+      SinkTaskContext context,
+      ErrantRecordReporter reporter
+  ) {
+    this(tp, storage, writerProvider, partitioner, connectorConfig, context, SYSTEM_TIME, reporter);
   }
 
   // Visible for testing
-  TopicPartitionWriter(TopicPartition tp,
-                       S3Storage storage,
-                       RecordWriterProvider<S3SinkConnectorConfig> writerProvider,
-                       Partitioner<?> partitioner,
-                       S3SinkConnectorConfig connectorConfig,
-                       SinkTaskContext context,
-                       Time time) {
+  TopicPartitionWriter(
+      TopicPartition tp,
+      S3Storage storage,
+      RecordWriterProvider<S3SinkConnectorConfig> writerProvider,
+      Partitioner<?> partitioner,
+      S3SinkConnectorConfig connectorConfig,
+      SinkTaskContext context,
+      Time time,
+      ErrantRecordReporter reporter
+  ) {
     this.connectorConfig = connectorConfig;
     this.time = time;
     this.tp = tp;
@@ -116,6 +124,7 @@ public class TopicPartitionWriter {
     this.context = context;
     this.writerProvider = writerProvider;
     this.partitioner = partitioner;
+    this.reporter = reporter;
     this.timestampExtractor = partitioner instanceof TimeBasedPartitioner
                                   ? ((TimeBasedPartitioner) partitioner).getTimestampExtractor()
                                   : null;
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -106,15 +106,14 @@ public class TopicPartitionWriter {
   }
 
   // Visible for testing
-  TopicPartitionWriter(
-      TopicPartition tp,
-      S3Storage storage,
-      RecordWriterProvider<S3SinkConnectorConfig> writerProvider,
-      Partitioner<?> partitioner,
-      S3SinkConnectorConfig connectorConfig,
-      SinkTaskContext context,
-      Time time,
-      ErrantRecordReporter reporter
+  TopicPartitionWriter(TopicPartition tp,
+                       S3Storage storage,
+                       RecordWriterProvider<S3SinkConnectorConfig> writerProvider,
+                       Partitioner<?> partitioner,
+                       S3SinkConnectorConfig connectorConfig,
+                       SinkTaskContext context,
+                       Time time,
+                       ErrantRecordReporter reporter
   ) {
     this.connectorConfig = connectorConfig;
     this.time = time;
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -195,8 +195,15 @@ public class TopicPartitionWriter {
     while (!buffer.isEmpty()) {
       try {
         executeState(now);
-      } catch (SchemaProjectorException | IllegalWorkerStateException e) {
+      } catch (IllegalWorkerStateException e) {
         throw new ConnectException(e);
+      } catch (SchemaProjectorException e) {
+        if (reporter != null) {
+          reporter.report(buffer.poll(), e);
+          log.warn("Errant record written to DLQ due to: {}", e.getMessage());
+        } else {
+          throw e;
+        }
       }
     }
     commitOnTimeIfNoData(now);
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -219,10 +219,6 @@ public class TopicPartitionWriter {
         // fallthrough
       case WRITE_PARTITION_PAUSED:
         SinkRecord record = buffer.peek();
-        if (recordProducesDataException(record)) {
-          buffer.poll(); // remove faulty record
-          break;
-        }
         if (timestampExtractor != null) {
           currentTimestamp = timestampExtractor.extract(record, now);
           if (baseRecordTimestamp == null) {
@@ -286,13 +282,12 @@ public class TopicPartitionWriter {
       setNextScheduledRotation();
       nextState();
     } else {
-      currentEncodedPartition = encodedPartition;
       SinkRecord projectedRecord = compatibility.project(
           record,
           null,
           currentValueSchema
       );
-      writeRecord(projectedRecord);
+      writeRecord(projectedRecord, encodedPartition);
       buffer.poll();
       if (rotateOnSize()) {
         log.info(
@@ -307,30 +302,6 @@ public class TopicPartitionWriter {
       }
     }
     return true;
-  }
-
-  /**
-   * Use a temporary writer to check the record for DataExceptions ahead of writing.
-   *
-   * @param record the record intended to write
-   * @return whether the record produced a DataException and should be skipped
-   */
-  private boolean recordProducesDataException(SinkRecord record) {
-    RecordWriter tempWriter = writerProvider.getRecordWriter(connectorConfig, "");
-    try {
-      tempWriter.write(record);
-      return false;
-    } catch (DataException e) {
-      if (reporter != null) {
-        reporter.report(record, e);
-        log.warn("Errant record written to DLQ due to: {}", e.getMessage());
-        return true;
-      } else {
-        throw new ConnectException(e);
-      }
-    } finally {
-      tempWriter.close();
-    }
   }
 
   private void commitOnTimeIfNoData(long now) {
@@ -477,11 +448,8 @@ public class TopicPartitionWriter {
     context.resume(tp);
   }
 
-  private RecordWriter getWriter(SinkRecord record, String encodedPartition)
+  private RecordWriter newWriter(SinkRecord record, String encodedPartition)
       throws ConnectException {
-    if (writers.containsKey(encodedPartition)) {
-      return writers.get(encodedPartition);
-    }
     String commitFilename = getCommitFilename(encodedPartition);
     log.debug(
         "Creating new writer encodedPartition='{}' filename='{}'",
@@ -523,10 +491,56 @@ public class TopicPartitionWriter {
     return fileKey(topicsDir, dirPrefix, name);
   }
 
-  private void writeRecord(SinkRecord record) {
-    currentOffset = record.kafkaOffset();
+  private void writeRecord(SinkRecord record, String encodedPartition) {
+    RecordWriter writer = writers.get(encodedPartition);
+    long currentOffsetIfSuccessful = record.kafkaOffset();
+    boolean shouldRemoveWriter = false;
+    boolean shouldRemoveStartOffset = false;
+    boolean shouldRemoveCommitFilename = false;
+    try {
+      if (!startOffsets.containsKey(encodedPartition)) {
+        log.trace(
+            "Setting writer's start offset for '{}' to {}",
+            encodedPartition,
+            currentOffsetIfSuccessful
+        );
+        startOffsets.put(encodedPartition, currentOffsetIfSuccessful);
+        shouldRemoveStartOffset = true;
+      }
+      if (writer == null) {
+        if (!commitFiles.containsKey(encodedPartition)) {
+          shouldRemoveCommitFilename = true;
+        }
+        writer = newWriter(record, encodedPartition);
+        shouldRemoveWriter = true;
+      }
+      // potentially revert:
+      // * startOffsets
+      // * commitFilenames
+      // * writers
+      writer.write(record);
+    } catch (DataException e) {
+      if (reporter != null) {
+        if (shouldRemoveStartOffset) {
+          startOffsets.remove(encodedPartition);
+        }
+        if (shouldRemoveWriter) {
+          writers.remove(encodedPartition);
+        }
+        if (shouldRemoveCommitFilename) {
+          commitFiles.remove(encodedPartition);
+        }
+        reporter.report(record, e);
+        log.warn("Errant record written to DLQ due to: {}", e.getMessage());
+        return;
+      } else {
+        throw new ConnectException(e);
+      }
+    }
 
-    if (!startOffsets.containsKey(currentEncodedPartition)) {
+    currentEncodedPartition = encodedPartition;
+    currentOffset = record.kafkaOffset();
+    if (shouldRemoveStartOffset) {
       log.trace(
           "Setting writer's start offset for '{}' to {}",
           currentEncodedPartition,
@@ -537,13 +551,9 @@ public class TopicPartitionWriter {
       // value, we know that we have at least one record. This allows us
       // to initialize all our maps at the same time, and saves future
       // checks on the existence of keys
-      startOffsets.put(currentEncodedPartition, currentOffset);
       recordCounts.put(currentEncodedPartition, 0L);
       endOffsets.put(currentEncodedPartition, 0L);
     }
-
-    RecordWriter writer = getWriter(record, currentEncodedPartition);
-    writer.write(record);
     ++recordCount;
 
     recordCounts.put(currentEncodedPartition, recordCounts.get(currentEncodedPartition) + 1);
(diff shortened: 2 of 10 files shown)


dosvath:
Build pending release of the [errantRecordReporter changes](https://github.com/confluentinc/kafka-connect-storage-common/commit/3b19abe6768826f20d95b95d9e8b83c7539203b3) in kafka-connect-storage-common.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -250,43 +261,61 @@ private boolean checkRotationOrAppend(
...
+        } else {
+          return false;
+        }
+      }
+      return true;
+    } catch (SchemaProjectorException e) {
+      if (reporter != null) {
+        reporter.report(record, e);
+        log.warn("Errant record written to DLQ due to SchemaProjectorException.");
+        buffer.poll();
```
dosvath, line 306: The tests using mock S3 fail if this buffer.poll() is not present. They fail on attempts to write records after the errant record with the exception: 
```
org.apache.avro.AvroRuntimeException: already open

	at org.apache.avro.file.DataFileWriter.assertNotOpen(DataFileWriter.java:93)
	at org.apache.avro.file.DataFileWriter.setCodec(DataFileWriter.java:101)
	at io.confluent.connect.s3.format.avro.AvroRecordWriterProvider$1.write(AvroRecordWriterProvider.java:75)
	at io.confluent.connect.s3.TopicPartitionWriter.writeRecord(TopicPartitionWriter.java:530)
	at io.confluent.connect.s3.TopicPartitionWriter.checkRotationOrAppend(TopicPartitionWriter.java:287)
	at io.confluent.connect.s3.TopicPartitionWriter.executeState(TopicPartitionWriter.java:231)
	at io.confluent.connect.s3.TopicPartitionWriter.write(TopicPartitionWriter.java:200)
	at io.confluent.connect.s3.TopicPartitionWriterTest.testSchemaProjectionExceptionReported(TopicPartitionWriterTest.java:418)
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -250,43 +261,61 @@ private boolean checkRotationOrAppend(
...
+      } else {
+        throw new ConnectException(e);
+      }
+    } catch (ConnectException e) {
+      // ConnectExceptions may come from the RecordWriterProviders on write
+      if (reporter != null) {
+        reporter.report(record, e);
+        log.warn("Errant record written to DLQ due to ConnectException.");
       }
+      throw e;
```
dosvath, line 317: propagating here would keep the current connector behavior, TBD is this is desired

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -250,43 +261,61 @@ private boolean checkRotationOrAppend(
...
+    } catch (SchemaProjectorException e) {
+      if (reporter != null) {
+        reporter.report(record, e);
+        log.warn("Errant record written to DLQ due to SchemaProjectorException.");
+        buffer.poll();
+        return true;
+      } else {
+        throw new ConnectException(e);
+      }
+    } catch (ConnectException e) {
```
dosvath, line 311: It may be useful to report the exceptions due to the keys/headers not being present when expected. These are thrown in [the KeyValueHeaderRecordWriterProvider](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/f42294dae9f94f0cb258f05a3ec7b3b9c664652d/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java#L91)

kkonstantine (commented):
Thanks @dosvath 
I've reviews only the sink task and topic partition writer code, where the reporter changes reside. 

We need to focus around `DataException` that may come from `storage-common` or from SR and `AvroData` (e.g. `avroData#fromConnectData`) and not only `SchemaProjectorException`. 

`DataException` when multipart upload fails most probably needs to change to `ConnectException`. Finally, let's review any other possible exception that needs to be caught and wrapped into `DataException` if such a record needs to go to the DLQ.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -177,6 +216,9 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
       TopicPartition tp = new TopicPartition(topic, partition);
 
       if (maybeSkipOnNullValue(record)) {
+        if (reporter != null) {
+          reporter.report(record, new ConnectException("Cannot write null value record"));
```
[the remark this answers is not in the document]
dosvath: Correct, a `DataException` will be thrown and we will handle the DLQ writing in the `TopicPartitionWriter`.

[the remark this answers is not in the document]
dosvath: Updated.

dosvath:
Thanks for the review @kkonstantine, I updated the PR. I looked into the possible `DataExceptions` in `AvroData` and it looks like they are propagated by the `AvroRecordWriterProvider`. I updated the `ByteArrayRecordWriterProvider` to propagate its `DataException` as well. These all seem appropriate to be reported in the DLQ.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
dosvath: correct

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -204,6 +215,9 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
...
```
dosvath, line 219: Should this be `DataException`?

kkonstantine (commented):
Leaving a comment that I had pending and I'll return for a another review.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
kkonstantine: I believe so. This would match the other use case right?
Have you checked how the exception is used by the reporter?

dosvath: Yes, this would match the other cases, updated.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -217,7 +217,9 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
 
       if (maybeSkipOnNullValue(record)) {
         if (reporter != null) {
-          reporter.report(record, new DataException("Cannot write null value record"));
+          String nullValueMsg = "Cannot write null value record";
+          reporter.report(record, new DataException(nullValueMsg));
+          log.warn("Errant record written to DLQ due to: {}", nullValueMsg);
```
kkonstantine, line 222: did we have a warning previously? This might generate too many messages if a user is skipping them (happens to have lot's of tombstones that you need to skip).

dosvath: We did not have a warning previously. There is also one in the TopicPartitionWriter for when a DataException is caught, but that one may not be as frequent so I think we may leave the other one.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -108,13 +110,23 @@ public void start(Map<String, String> props) {
...
 
       open(context.assignment());
+      try {
+        reporter = context.errantRecordReporter();
+        if (reporter == null) {
+          log.info("Errant record reporter not configured.");
+        }
+      } catch (NoSuchMethodError | NoClassDefFoundError | UnsupportedOperationException e) {
+        // Will occur in Connect runtimes earlier than 2.6
+        log.warn("AK versions prior to 2.6 do not support the errant record reporter");
```
kkonstantine, line 127:
```suggestion
        log.warn("Connect versions prior to Apache Kafka 2.6 do not support the errant record reporter");
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -498,8 +517,19 @@ private void writeRecord(SinkRecord record) {
...
-    writer.write(record);
+    RecordWriter writer = getWriter(currentEncodedPartition);
+    try {
+      writer.write(record);
+    } catch (DataException e) {
+      if (reporter != null) {
+        reporter.report(record, e);
+        log.warn("Errant record written to DLQ due to: {}", e.getMessage());
+        writer.close();
+        writers.remove(currentEncodedPartition); // writer keeps exception state otherwise
```
kkonstantine, line 528: the comment doesn't read very well for me. 
Does this work for cases where the record is both first or somewhere in the middle of the currentEncodedPartiion?

kkonstantine (commented):
Thanks @dosvath 
Couple more questions. Seems we are close

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -498,8 +517,19 @@ private void writeRecord(SinkRecord record) {
...
-    RecordWriter writer = getWriter(record, currentEncodedPartition);
-    writer.write(record);
+    RecordWriter writer = getWriter(currentEncodedPartition);
+    try {
+      writer.write(record);
+    } catch (DataException e) {
+      if (reporter != null) {
+        reporter.report(record, e);
+        log.warn("Errant record written to DLQ due to: {}", e.getMessage());
+        writer.close();
```
kkonstantine, line 527: I know see we can't close here and advance the offsets. We'll have to keep the writer

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
dosvath: Updated, this is N/A anymore.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
dosvath: Updated, the actual writer won't be closed.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -204,6 +217,9 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
       TopicPartition tp = new TopicPartition(topic, partition);
 
       if (maybeSkipOnNullValue(record)) {
+        if (reporter != null) {
+          reporter.report(record, new DataException("Cannot write null value record"));
```
levzem, line 221:
```suggestion
          reporter.report(record, new DataException("Skipping null value record."));
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -186,8 +196,15 @@ public void write() {
...
-      } catch (SchemaProjectorException | IllegalWorkerStateException e) {
+      } catch (IllegalWorkerStateException e) {
         throw new ConnectException(e);
+      } catch (SchemaProjectorException e) {
+        if (reporter != null) {
+          reporter.report(buffer.poll(), e);
+          log.warn("Errant record written to DLQ due to: {}", e.getMessage());
+        } else {
+          throw e;
+        }
```
levzem, line 207: do we want to skip the record if we report it

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -288,6 +309,30 @@ private boolean checkRotationOrAppend(
     return true;
   }
 
+  /**
+   * Use a temporary writer to check the record for DataExceptions ahead of writing.
```
levzem, line 313: do we need to check ahead? seems expensive

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -169,7 +169,9 @@ public void commit() throws IOException {
       log.debug("Upload complete for bucket '{}' key '{}'", bucket, key);
     } catch (IOException e) {
       log.error("Multipart upload failed to complete for bucket '{}' key '{}'", bucket, key);
-      throw new DataException("Multipart upload failed to complete.", e);
+      throw new RetriableException(
+          String.format("Multipart upload failed to complete: %s", e.getMessage())
+      );
```
levzem, line 174: lets be careful here that we don't overwrite this PR #378

not sure we want to retry this. keep this as `DataException`

this is in the `commit()`, shouldn't affect `write()`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
levzem: `DataException` is thrown by [fromConnectData()](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/349/files#diff-16f592a34da31ec963db3a13a0726eeb8b9d6176475ebe05ba84d1c760aa3ff4R71-R72 ), so I don't think we need this check ahead. just try to [write here](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/349/files#diff-4e620d031f99e2962510e761f99cae963938454ad0e71a6a79d1b1cf3c7ffc99L501) and catch the data exception

levzem (commented):
some high level comments

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
dosvath: I can't open you diff links for some reason, let's try file links. Overall I agree with that we shouldn't have a check ahead and I tried not to have it. I believe you mention the existing position of the write, the reason I wasn't able to handle this there is because the logic before that write call affects global vars and we'd have to manually restore those states after the catch, which seemed very complex. This seemed the most readable solution, lmk if you have ideas.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
dosvath: The reason I changed this was based on a previous discussion with @kkonstantine that we should only have a `DataException` when the data has an issue. This looked more like an issue that is not caused by the data in the record itself. The approach may have changed since your PR so let's double check.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
dosvath: We'd want to skip writing the record to the S3 file if it's been reported, is that what you mean?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
dosvath: Instead of DataException, we could also use an exception class that is not retriable but more appropriate for the cause?

levzem: why not a `ConnectException` then?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
levzem: as long as customers understand that enabling the reporter implicitly subscribes them to skipping records instead of failing

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
levzem: just write the record first, and if it succeeds then change all of the global variables. should be pretty simple

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
dosvath: That works for me 👍 The naming `testPropagateRetriableS3Exceptions` from PR #378 tests made me think we consider it retriable.

levzem: this case will fail the connector so lets keep it that way. That test tests something else.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
dosvath: The current `TopicPartitionWriter` implementation causes issues with that approach:
1. The writer used to write the record is itself retrieved based on the configuration of the global vars, so this doesn't allow us to write the record ahead of setting the global vars. 

    The writer ahead of the write is retrieved on [L#501](https://github.com/dosvath/kafka-connect-storage-cloud/blob/4cd601138a838f61fba6dbd4f314a470c01c7dac/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L501). If you trace through the method calls made by that retrieval you will see one of the methods used `getCommitFilename` uses the global var  `startOffsets`.

2. Additional global variables are set at a higher level ahead of the [writeRecord](https://github.com/dosvath/kafka-connect-storage-cloud/blob/4cd601138a838f61fba6dbd4f314a470c01c7dac/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L275) call. The write call is dependent on these to be set ahead. [currentEncodedPartition here](https://github.com/dosvath/kafka-connect-storage-cloud/blob/4cd601138a838f61fba6dbd4f314a470c01c7dac/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L269), [currentSchemas here](https://github.com/dosvath/kafka-connect-storage-cloud/blob/4cd601138a838f61fba6dbd4f314a470c01c7dac/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L216). 
Setting these ahead of the write is currently required for the `TopicPartitionWriter` to function properly. 

Refactoring the `TopicPartitionWriter` class to behave differently can be done but would first require a major refactoring PR without any DLQ changes, and would need to be thoroughly tested to eliminate possible regressions.

dosvath: If it's any consolation, keep in mind the above temporary writer does not write immediately to a file, only if commit is called, and if it's a correct record the result will be cleaned up by JVM when the tmp writer is closed.

levzem: I will look more into this, but does writing a bad record, pollute the writer? any chance we could just try writing, fail, and keep moving?

dosvath: I believe the writer itself wouldn't be the issue in that case, it could be used for further writing, but the record count in the file would not align with the offsets in the TPW class and cause misalignment in the globals mentioned, resulting in failing unit tests/ altered behavior of the class.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -288,6 +309,30 @@ private boolean checkRotationOrAppend(
...
+    RecordWriter tempWriter = writerProvider.getRecordWriter(connectorConfig, "");
+    try {
+      tempWriter.write(record);
+      tempWriter.close();
+      return false;
+    } catch (DataException e) {
+      if (reporter != null) {
+        reporter.report(record, e);
+        log.warn("Errant record written to DLQ due to: {}", e.getMessage());
+        tempWriter.close();
```
levzem, line 328: can put closing the tempWriter in a `finally`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -108,13 +110,24 @@ public void start(Map<String, String> props) {
...
       writerProvider = newRecordWriterProvider(connectorConfig);
       partitioner = newPartitioner(connectorConfig);
 
       open(context.assignment());
+      try {
+        reporter = context.errantRecordReporter();
+        if (reporter == null) {
+          log.info("Errant record reporter not configured.");
+        }
+      } catch (NoSuchMethodError | NoClassDefFoundError | UnsupportedOperationException e) {
```
levzem, line 125:
```suggestion
      } catch (NoSuchMethodError | NoClassDefFoundError e) {
```

we don't need this IIUC

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/S3SinkConnectorIT.java
```diff
@@ -193,37 +222,160 @@ private void testBasicRecordsWritten(String expectedFileExtension) throws Throwa
...
+
+  private void produceRecordsWithHeadersNoKey(String topic, int recordCount, SinkRecord record) throws Exception {
+    // Send records to Kafka
+    for (long i = 0; i < recordCount; i++) {
+      byte[] kafkaValue = jsonConverter.fromConnectData(record.topic(), record.valueSchema(), record.value());
+      ProducerRecord<byte[],byte[]> producerRecord =
+          new ProducerRecord<byte[],byte[]>(topic, TOPIC_PARTITION, null, kafkaValue, sampleHeaders());
+      producer.send(producerRecord).get();
+    }
+  }
```
levzem, line 351: use booleans instead to see if to produce with a key/value/header

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/S3SinkConnectorIT.java
```diff
@@ -466,6 +619,39 @@ private void initializeJsonConverter() {
     jsonConverter.configure(jsonConverterProps);
   }
 
+  // whether a filename contains any of the extensions
+  private boolean filenameContainsExtensions(String filename, List<String> extensions) {
+    for (String extension : extensions){
+      if (filename.contains(extension)) {
```
levzem, line 625:
```suggestion
      if (filename.endsWith(extension)) {
```

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/S3SinkConnectorIT.java
```diff
@@ -466,6 +619,39 @@ private void initializeJsonConverter() {
...
+  private List<String> getS3FileListValues(List<S3ObjectSummary> summaries) {
+    List<String> excludeExtensions = Arrays.asList(".headers.avro", ".keys.avro");
+    List<String> filteredFiles = new ArrayList<>();
+    for (S3ObjectSummary summary : summaries) {
+      String fileKey = summary.getKey();
+      if (!filenameContainsExtensions(fileKey, excludeExtensions)) {
+        filteredFiles.add(fileKey);
+      }
+    }
+    return filteredFiles;
```
levzem, line 642: could use a stream for this

levzem (commented):
@dosvath you were right, i tried getting it to work, but the problem was that some of the writers actually get polluted by writing a bad record so we will have to go with this approach unless we do a significant refactor

levzem:
> @dosvath you were right, i tried getting it to work, but the problem was that some of the writers actually get polluted by writing a bad record so we will have to go with this approach unless we do a significant refactor

did some more thinking and if we change the `RecordWriter` interface and add an `isValidRecord()` method, we can have this be a lot nicer because it will help extend DLQ to all storage connectors because we will have this issue everywhere i.e HDFS. Each record writer can decide if the record is valid or not (most of them just do `toConnectSchema()` or `toConnectData()`) and this will be better than just creating temporary record writers

dosvath:
> > @dosvath you were right, i tried getting it to work, but the problem was that some of the writers actually get polluted by writing a bad record so we will have to go with this approach unless we do a significant refactor
> 
> did some more thinking and if we change the `RecordWriter` interface and add an `isValidRecord()` method, we can have this be a lot nicer because it will help extend DLQ to all storage connectors because we will have this issue everywhere i.e HDFS. Each record writer can decide if the record is valid or not (most of them just do `toConnectSchema()` or `toConnectData()`) and this will be better than just creating temporary record writers

That would be a nice solution as well. Additionally, if we want to save some computation on the check, we may only need to call `isValidRecord()` when we have a reporter enabled, since in the other case we fail with the exception.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
dosvath: You're right that for the reporter the `UnsupportedOperationException`  is not needed. This is needed in storage connectors that use the `StorageSinkTestBase` class in tests, thrown [here](https://github.com/confluentinc/kafka-connect-storage-common/blob/e1aa07bdbcd031d8f26a577a7bf3984b4a9ca243/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java#L254). The class is used in `S3SinkTaskTest`.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
kkonstantine: Unfortunately we can't introduce a performance regression per record if anyone enables error reporting. This performance tax is not an option. That's the bad news. 

The good news is that error inspection happens per record. So we can think like a compiler, meaning we can do register allocation (move variables around in local copies if we have to) and instruction reordering, keeping the instructions that need to happen before `writer.write(record)` before (in which case we'll track and revert the state they are changing if an error happens), while leaving any other instructions that mutate the state of the TPW after, to be executed only if a write succeeds.

kkonstantine (commented):
Thanks for addressing the comments so far @dosvath

I think we are close. One last effort that it's worth undertaking is to handle the `DataException` within the TPW. I pushed a draft commit to give you an idea how this could be done. We'll need to fix tests and corner cases for when the issue surfaces earlier in the schema projection. 

Let me know what you think. cc @levzem

dosvath:
> Thanks for addressing the comments so far @dosvath
> 
> I think we are close. One last effort that it's worth undertaking is to handle the `DataException` within the TPW. I pushed a draft commit to give you an idea how this could be done. We'll need to fix tests and corner cases for when the issue surfaces earlier in the schema projection.
> 
> Let me know what you think. cc @levzem

Thanks @kkonstantine. 
I updated the PR, one conflict remains in the TPW I noted the conflict it in the comments [here (1)](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/349/files#diff-4e620d031f99e2962510e761f99cae963938454ad0e71a6a79d1b1cf3c7ffc99R276) and [here (2)](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/349/files#diff-4e620d031f99e2962510e761f99cae963938454ad0e71a6a79d1b1cf3c7ffc99R296). We should take a look what approach would be best here, @levzem also take a look if you can.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -915,78 +923,100 @@ public void testAddingS3ObjectTags() throws Exception{
...
+    testExceptionReportedToDLQ(faultyRecord, DataException.class, exceptionMessage, true);
   }
 
+  // test DLQ for lower level exception coming from AvroData
   @Test
-  public void testRecordKeysAnsHeadersWritten() throws Exception {
-    writeRecordWithKeysAndHeaders(null);
+  public void testDataExceptionReportedNullValueForNonOptionalSchema() throws Exception {
+    SinkRecord faultyRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, "key",
+        Schema.STRING_SCHEMA, null, 1, 0L, TimestampType.NO_TIMESTAMP_TYPE, sampleHeaders());
```
kkonstantine, line 997: This schema selection is incorrect. 

The schema differs from the other records, and will cause a rotation due to schema compatibility checks. But I don't think this is what you want to check here. See if the suggestion below makes sense: 

```suggestion
        createSchema(), null, 1, 0L, TimestampType.NO_TIMESTAMP_TYPE, sampleHeaders());
```

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -915,78 +923,100 @@ public void testAddingS3ObjectTags() throws Exception{
...
+    testExceptionReportedToDLQ(faultyRecord, SchemaProjectorException.class, exceptionMessage, false);
+  }
+
+  // test DLQ for lower level exception coming from AvroData
+  @Test
+  public void testDataExceptionReportedIncorrectSchema() throws Exception {
+    Schema sampleValueSchema = createSchema();
+    int kafkaOffset = 1;
+    SinkRecord faultyRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, "key",
+        sampleValueSchema, "1", kafkaOffset, 0L, TimestampType.NO_TIMESTAMP_TYPE, sampleHeaders());
```
kkonstantine, line 984:
```suggestion
    int kafkaOffset = 1;
    SinkRecord faultyRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, "key",
        createSchema(), "1", kafkaOffset, 0L, TimestampType.NO_TIMESTAMP_TYPE, sampleHeaders());
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -478,10 +505,56 @@ private String fileKeyToCommit(String dirPrefix, long startOffset) {
...
+        if (!commitFiles.containsKey(encodedPartition)) {
+          shouldRemoveCommitFilename = true;
+        }
+        writer = newWriter(record, encodedPartition);
+        shouldRemoveWriter = true;
+      }
+      // potentially revert:
+      // * startOffsets
+      // * commitFilenames
+      // * writers
```
kkonstantine, line 534:
```suggestion
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -249,43 +266,56 @@ private boolean checkRotationOrAppend(
       String encodedPartition,
       long now
   ) {
+    // rotateOnTime check must go before writeRecord.
```
kkonstantine, line 269:
```suggestion
    // rotateOnTime is safe to go before writeRecord, because it is acceptable even for a faulty record to trigger time-based rotation if it applies 
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -249,43 +266,56 @@ private boolean checkRotationOrAppend(
...
+      setNextScheduledRotation();
+      nextState();
+      return true;
+    }
+
+    // Conflict 1:
+    // If shouldChangeSchema is true, a rotation must be performed before writeRecord
+    // to avoid mismatching schemas in the writer.
+    // Additionally, this check must happen ahead of writeRecord,
+    // otherwise a SchemaProjectionException may be missed.
```
kkonstantine, line 280: It's sufficient to assume that faults are independent of schema mismatches. If both happen in the user's code then it's fine for the rotation to happen. `null` keys and empty headers are not allowed when key and header export is enabled. 
```suggestion
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -249,43 +266,56 @@ private boolean checkRotationOrAppend(
...
-          record,
-          null,
-          currentValueSchema
+      return true;
+    }
+
+    // Conflict 2:
+    // A writeRecord attempt must be before the shouldChangeSchema file rotation behavior so
+    // a faulty record does not cause a rotation before getting to writeRecord
+    // where we would get the DataException.
```
kkonstantine, line 299: Schema compatibility takes into account only the value schema at the moment
```suggestion
```

kkonstantine (commented):
Suggested a fix in one test. 
The tests on null keys and empty headers needed to be adjusted to account for file rotation.

kkonstantine (approved):
Huge improvement to the connector's capability to cope with undesired poison pill records. 
Thank you @dosvath 

LGTM. 

Let's just make sure it's targeting the `10.0.x` branch. That's where this change should go

kafka-connect-s3/pom.xml
```diff
@@ -21,7 +21,7 @@
     <parent>
         <groupId>io.confluent</groupId>
         <artifactId>kafka-connect-storage-cloud</artifactId>
-        <version>10.0.0-SNAPSHOT</version>
+        <version>10.1.0-SNAPSHOT</version>
```
kkonstantine, line 24: We should not make this change. We are still targeting to release 10.0.0

pom.xml
```diff
@@ -27,7 +27,7 @@
     <groupId>io.confluent</groupId>
     <artifactId>kafka-connect-storage-cloud</artifactId>
     <packaging>pom</packaging>
-    <version>10.0.0-SNAPSHOT</version>
```
kkonstantine, line 30: The changes here are not needed

PR #351, opened 2020-08-14
CC-11250: Add basic integration tests

dosvath:
## Summary
This PR adds basic integration tests to the S3 sink connector. These tests are an extension to our internal muckrake system test framework, and ITs may be added here to be used as an alternative for future functionalities that are not covered by muckrake.

Note: These tests do not yet have complete muckrake test coverage. This PR is meant to add the basic coverage for the S3 sink connector, for verifying the naming, counts, and contents of files written to an S3 bucket. The tests include cases for `.avro`, `.parquet` and `.json` file encodings. The PR is scoped to the basic test cases for review convenience, and more complex test cases may be added in follow-up PRs. As new functionality PRs are made to S3, integration tests can now be included along with the functionality changes. 

## New Classes
**BaseConnectorIT**
The standard `BaseConnectorIT` class from most connectors, with added utility functions for S3. 

**ConnectProducer**
A class that enables sending SinkRecords to the EmbeddedConnectCluster via a custom Producer. This class will be useful for future tests as it allows highly customizable SinkRecords with structs, and null values for negative path tests.  

**S3SinkConnectorIT**
The main integration test class for S3 that contains the individual test cases.

- [ ] yes
- [X] no

## Release Plan

 
Backporting branch TBD, feel free to suggest in comments.


levzem:
@dosvath what about using the [S3 testing mock ](https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/kafka-connect-s3-source/src/test/java/io/confluent/connect/s3/source/S3MockStorage.java#L28-L37) so that we can run these tests in Jenkins

dosvath:
> @dosvath what about using the [S3 testing mock ](https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/kafka-connect-s3-source/src/test/java/io/confluent/connect/s3/source/S3MockStorage.java#L28-L37) so that we can run these tests in Jenkins

@levzem the current tests use S3 mock, the goal of these integration tests is to use a real S3 bucket, and get as close to real world scenarios as possible. For Jenkins, we will be having an initiative to make these tests, and tests for other connectors, runnable. The solution will involve setting up credentials access on Jenkins. cc @sonupillai

levzem:
> @levzem the current tests use S3 mock, the goal of these integration tests is to use a real S3 bucket, and get as close to real world scenarios as possible. For Jenkins, we will be having an initiative to make these tests, and tests for other connectors, runnable. The solution will involve setting up credentials access on Jenkins. cc @sonupillai

well we can write these ITs to use the mock if it detects no S3 credentials or use a real S3 bucket if it does (or based on some maven property) in the meantime.

dosvath:
> > @levzem the current tests use S3 mock, the goal of these integration tests is to use a real S3 bucket, and get as close to real world scenarios as possible. For Jenkins, we will be having an initiative to make these tests, and tests for other connectors, runnable. The solution will involve setting up credentials access on Jenkins. cc @sonupillai
> 
> well we can write these ITs to use the mock if it detects no S3 credentials or use a real S3 bucket if it does (or based on some maven property) in the meantime.

Good thought, though the ITs in this PR use the `EmbeddedConnectCluster` which start the sink connector from property configs. [See here](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/351/files#diff-81685e186f388ca0dcddaa72de3fa739R99). This is part of the advantage of the ITs, they give us an actual connector instance in the embedded kafka cluster, vs the current tests in the repo only start a connector task. We reach the issue when you look at how we are providing the S3 mock instance to the existing tests, the mock S3 object is wrapped in a `S3Storage` instance and passed directly to the connector task. This is not possible in the `EmbeddedCluster` type tests, because we don't have access to the task, we set up the connector in the cluster from the properties map.

levzem:
@dosvath we set the url `https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/kafka-connect-s3-source/src/test/java/io/confluent/connect/s3/source/S3MockStorage.java#L34` and then can pass that url as a config to the connector. As you can see, this code is actually running in an IT so it is entirely possible :)

dosvath:
> @dosvath we set the [url here](https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/kafka-connect-s3-source/src/test/java/io/confluent/connect/s3/source/S3MockStorage.java#L34) and then can pass that url as a config to the connector. As you can see, this code is actually [running in an IT](https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/kafka-connect-s3-source/src/test/java/io/confluent/connect/s3/source/integration/S3SourceConnectorIT.java#L29) so it is entirely possible :)

Cool, that brings us closer to the possibility, now only a few questions remain. How do we specify the URL to the connector config map? `s3.proxy.url` I believe is for another purpose? Also the issue of S3 looking for credentials remains.

dosvath:
> @dosvath we set the [url here](https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/kafka-connect-s3-source/src/test/java/io/confluent/connect/s3/source/S3MockStorage.java#L34) and then can pass that url as a config to the connector. As you can see, this code is actually [running in an IT](https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/kafka-connect-s3-source/src/test/java/io/confluent/connect/s3/source/integration/S3SourceConnectorIT.java#L29) so it is entirely possible :)

That's great. I'll give it a try. One question, what about when the S3 connector looks for credentials?

levzem:
@dosvath `store.url` should be the config for that, as for credentials, you can specify the credentials to just accept anything as the links I have attached.

dosvath:
Thanks @levzem, I added the functionality to use the mock client when credentials aren't available. It's ready for another pass.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -0,0 +1,483 @@
...
+    clearBucket(TEST_BUCKET_NAME);
+  }
+
+  @After
+  public void close() throws Throwable {
+    // stop all Connect, Kafka and Zk threads.
+    connect.stop();
+    // delete the downloaded test file folder
+    FileUtils.deleteDirectory(new File(TEST_DOWNLOAD_PATH));
+    clearBucket(TEST_BUCKET_NAME);
```
levzem, line 160: maybe we should delete the buckets esp if we are using actual S3. should fully clean up the test environment

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -0,0 +1,483 @@
...
+   * @return the rows of the file as JsonNodes
+   */
+  private List<JsonNode> getContentsFromAvro(String filePath) throws IOException {
+    DatumReader<GenericRecord> datumReader = new GenericDatumReader<>();
+    DataFileReader<GenericRecord> dataFileReader = new DataFileReader(new File(filePath),
+        datumReader);
+    List<JsonNode> fileRows = new ArrayList<>();
+    while (dataFileReader.hasNext()) {
+      GenericRecord row = dataFileReader.next();
+      log.debug("Avro row: {}", row);
```
levzem, line 398: not sure we need these logs

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -0,0 +1,483 @@
...
+  private List<JsonNode> getContentsFromParquet(String filePath) throws IOException {
+    ParquetReader<SimpleRecord> reader = ParquetReader
+        .builder(new SimpleReadSupport(), new Path(filePath)).build();
+    ParquetMetadata metadata = ParquetFileReader
+        .readFooter(new Configuration(), new Path(filePath));
+    JsonRecordFormatter.JsonGroupFormatter formatter = JsonRecordFormatter
+        .fromSchema(metadata.getFileMetaData().getSchema());
+    List<JsonNode> fileRows = new ArrayList<>();
+    for (SimpleRecord value = reader.read(); value != null; value = reader.read()) {
+      log.debug("Parquet row: {}", formatter.formatRecord(value));
```
levzem, line 420: same as above

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -0,0 +1,483 @@
...
+      return getContentsFromAvro(filePath);
+    } else if (filePath.endsWith(".parquet")) {
+      return getContentsFromParquet(filePath);
+    } else if (filePath.endsWith(".json")) {
+      return getContentsFromJson(filePath);
+    } else {
+      throw new InvalidFileTypeException(
+          String.format("Downloaded file has unsupported extension: %s", filePath)
+      );
+    }
```
levzem, line 382: if we wanna be fancy we can use a map of extensions to Function<> and call that like here https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/cloud-storage-source-common/src/test/java/io/confluent/connect/cloud/storage/source/integration/StorageSourceConnectorIT.java#L49-L54 and here https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/cloud-storage-source-common/src/test/java/io/confluent/connect/cloud/storage/source/integration/StorageSourceConnectorIT.java#L320-L328

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -0,0 +1,483 @@
...
+import org.apache.parquet.tools.read.SimpleRecord;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.experimental.categories.Category;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Category(IntegrationTest.class)
+public abstract class BaseConnectorIT {
```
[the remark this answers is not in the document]
dosvath: Would you be okay with leaving the methods that verify file contents in this class? I feel it would keep the S3IT test cleaner, leaving mainly the `@Test` methods in that class. I could also make a separate utility file for these S3 convenience methods if we should move them from the base connector class.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/ConnectProducer.java
```diff
@@ -0,0 +1,84 @@
+/*
```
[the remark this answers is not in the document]
dosvath: I believe we had to use this class in the [past](https://github.com/confluentinc/kafka-connect-gcp-spanner/blob/master/src/test/java/io/confluent/connect/gcp/spanner/integration/ConnectProducer.java) because we are producing SinkRecords. I like the approach you suggested and tried it, but turns out that `formatter.formatValue` only accepts SinkRecords, not source records as we need for this sink. Let me know if you have any other ideas.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
dosvath: If-else seems okay for simplicity, since we'd want the exception to be thrown if none match.

levzem: it can just throw NPE which is fine for ITs

levzem: also allows for extension of tests when we want to test compression which have different extensions

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/ConnectProducer.java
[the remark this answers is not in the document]
levzem: why do you need to produce SourceRecords... SourceRecords are for SourceConnectors. I have been using this approach in all Sink connectors and it has been working fine

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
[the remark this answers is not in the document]
levzem: I think it would make the most sense in the current S3 IT class

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/ConnectProducer.java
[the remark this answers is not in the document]
dosvath: Awesome, that worked. I was going off of a previous example...

dosvath:
Thanks for the suggestions @levzem, I was able to remove the custom producer class by using the `SinkRecord` produce approach with the formatter.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/S3SinkConnectorIT.java
```diff
@@ -0,0 +1,472 @@
...
+      String kafkaValue = new String(formatter.formatValue(record), UTF_8);
+      connect.kafka().produce(TEST_TOPIC_NAME, null, kafkaValue);
+    }
+
+    log.info("Waiting for files in S3...");
+    int expectedFileCount = (int) NUM_RECORDS_INSERT / FLUSH_SIZE_STANDARD;
+    waitForFilesInBucket(TEST_BUCKET_NAME, expectedFileCount);
+
+    assertTrue(fileNamesValid(TEST_BUCKET_NAME, TEST_TOPIC_NAME, EXPECTED_PARTITION, expectedFileExtension));
+    assertTrue(fileContentsAsExpected(TEST_BUCKET_NAME, FLUSH_SIZE_STANDARD, recordValueStruct));
```
levzem, line 198: would nest these under one function

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/S3SinkConnectorIT.java
```diff
@@ -0,0 +1,472 @@
...
+      if (!fileInfo.namingIsValid()
+          || !fileInfo.filenameTopic.equals(expectedTopic)
+          || fileInfo.directoryPartition != expectedPartition
+          || !fileInfo.extension.equals(expectedExtension)
+      ) {
+        return false;
+      }
+    }
+    return true;
+  }
```
levzem, line 291: probably should just generate a list of expected files depending on the flush size and partitioner

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/S3SinkConnectorIT.java
```diff
@@ -0,0 +1,535 @@
...
+      S3FileInfo fileInfo = new S3FileInfo(file.getKey());
+      if (!fileInfo.namingIsValid()
+          || !fileInfo.filenameTopic.equals(expectedTopic)
+          || fileInfo.directoryPartition != expectedPartition
+          || !fileInfo.extension.equals(expectedExtension)
+      ) {
+        return false;
+      }
+      actualFiles.add(file.getKey());
+    }
```
levzem, line 348: no reason to make this check if you already have the expected file names

levzem (commented):
almost there, mostly nits and 1 comment

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/S3SinkConnectorIT.java
dosvath: It would allow us to return sooner, may save some time in the upcoming network ITs that need to list 1000+ files.

levzem: it would not allow to return sooner. the list comparison will fail as soon as it finds one element that does not match, just like this code does. the added code and complexity also do not make it worth it. should only optimize if we need to

dosvath: sounds good

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/S3SinkConnectorIT.java
```diff
@@ -0,0 +1,524 @@
...
+  private boolean fileContentsAsExpected(String bucketName, int expectedRowsPerFile,
+      Struct expectedRow) {
+    log.info("expectedRow: {}", expectedRow);
+    for (S3ObjectSummary file : S3Client.listObjectsV2(bucketName).getObjectSummaries()) {
+      String destinationPath = TEST_DOWNLOAD_PATH + file.getKey();
+      File downloadedFile = new File(destinationPath);
+      log.info("Saving file to : {}", destinationPath);
+      S3Client.getObject(new GetObjectRequest(bucketName, file.getKey()), downloadedFile);
+
+      String fileExtension = new S3FileInfo(file.getKey()).extension;
```
levzem, line 362: should just extract the extension here so we can get rid of this class since its only being used for the extension now

levzem (commented):
last comment

levzem (approved):
LGTM

Issue #337, opened 2020-06-09
S3 virtual host-style access bucket compatibility?  


aliadnani:
I read from the [S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html) that AWS will be deprecating path style links and switching over to virtual hosted style URLs for bucket access from the **30th September 2020** onwards, i.e: 

```
# Path Style
https://s3.us-west-2.amazonaws.com/mybucket/puppy.jpg

# Virtual Hosted-Style
https://my-bucket.s3.us-west-2.amazonaws.com/puppy.png
```

Will this be a problem in the future for the S3 connector? - as from the source, I'm guessing that this connector uses path style access (correct me if I am wrong).

aliadnani:
I found out you can change the bucket access path style from path style to virtual-hosted style by changing `.withPathStyleAccessEnabled(true)` to **false** 

in the `AmazonS3 newS3Client` method under `kafka-connect-storage-cloud/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java`

and then recompiling.

Parameterizing this setting may be a good idea in the future :)

PR #356, opened 2020-08-27
CC-11480: Add configuration property to choose between path style and virtual hosted bucket access to S3

kkonstantine:
## Problem
The connector since the beginning has been explicitly choosing to use path style access for S3 buckets in its requests. This access type is being deprecated and removed in Sept 30, 2020 for _new_ buckets on S3 that are created after that date. 

Addresses issue: https://github.com/confluentinc/kafka-connect-storage-cloud/issues/337

## Solution
A configuration property `s3.path.style.access.enabled` is introduced to control the behavior. The behavior is matching the current existing behavior and will be backported to previous connector versions. A new connector version will be generated in a follow up that will have virtual hosted access as the default. 

- [x] yes
- [ ] no

##### If yes, where?
S3 source connector

## Test Strategy
End to end tests with buckets in several regions and testing with and without setting `store.url`. 

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [x] System tests
- [x] Manual tests

## Release Plan

 
See above

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
@@ -37,6 +37,7 @@ import io.confluent.connect.storage.Storage;
 import io.confluent.connect.storage.common.util.StringUtils;
 
 import static io.confluent.connect.s3.S3SinkConnectorConfig.REGION_CONFIG;
+import static io.confluent.connect.s3.S3SinkConnectorConfig.S3_PATH_STYLE_ACCESS_ENABLED_CONFIG;
 import static io.confluent.connect.s3.S3SinkConnectorConfig.S3_PROXY_URL_CONFIG;
 import static io.confluent.connect.s3.S3SinkConnectorConfig.S3_RETRY_BACKOFF_CONFIG;
 import static io.confluent.connect.s3.S3SinkConnectorConfig.S3_RETRY_MAX_BACKOFF_TIME_MS;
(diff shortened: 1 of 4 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java
```diff
@@ -166,7 +169,7 @@ public boolean exists(String name) {
   }
 
   public boolean bucketExists() {
-    return StringUtils.isNotBlank(bucketName) && s3.doesBucketExist(bucketName);
+    return StringUtils.isNotBlank(bucketName) && s3.doesBucketExistV2(bucketName);
```
dosvath, line 172: any preference for V2 over non V2?

kkonstantine: `doesBucketExist` is deprecated and shows up as a warning. 

Details in the AWS SDK about "V1" not being able to distinguish between bucket permissions and wrong credentials. 

https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Client.html#doesBucketExist-java.lang.String-

kkonstantine:
@levzem This connector has external system tests currently (not ITs).
Two preliminary runs can be found here: [4267](https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/4267/) and [4268](https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/4268/)
and the tests will be adapted to reflect the new config.

kkonstantine:
@rhauch definitely. Merging here to have this fix ready for the RCs and I'll follow up with the docs PR. 

Thanks all!

PR #364, opened 2020-09-24
MINOR: Remove unused prop from test (that is going away)

rayokota:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


kkonstantine (approved):
Thanks @rayokota 
I see this is targeting previous releases and does not fail the test without it. 
Do we need to coordinate the merge, or we are good to go?

LGTM

rayokota:
Thanks @kkonstantine , we should be good to go.

PR #365, opened 2020-09-25
CC-11945: Fix empty header check for storing headers

dosvath:
## Problem
Writing empty headers should also throw an exception when `store.kafka.headers` is on. 

## Solution
Add empty check along existing null check. 

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
@@ -95,7 +95,8 @@ public class KeyValueHeaderRecordWriterProvider
         }
 
         // headerWriter != null means writing headers is turned on
-        if (headerWriter != null && sinkRecord.headers() == null) {
+        if (headerWriter != null
+            && (sinkRecord.headers() == null || sinkRecord.headers().isEmpty())) {
           throw new DataException(
               String.format("Headers cannot be null for SinkRecord: %s", sinkRecord)
           );
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
@@ -86,21 +86,22 @@ public class KeyValueHeaderRecordWriterProvider
       public void write(SinkRecord sinkRecord) {
         valueWriter.write(sinkRecord); // null check happens in sink task
         // keyWriter != null means writing keys is turned on
-        if (keyWriter != null && sinkRecord.key() == null) {
-          throw new DataException(
-              String.format("Key cannot be null for SinkRecord: %s", sinkRecord)
-          );
-        } else if (keyWriter != null) {
+        if (keyWriter != null) {
+          if (sinkRecord.key() == null) {
+            throw new DataException(
+                String.format("Key cannot be null for SinkRecord: %s", sinkRecord)
+            );
+          }
           keyWriter.write(sinkRecord);
         }
 
         // headerWriter != null means writing headers is turned on
-        if (headerWriter != null
-            && (sinkRecord.headers() == null || sinkRecord.headers().isEmpty())) {
-          throw new DataException(
-              String.format("Headers cannot be null for SinkRecord: %s", sinkRecord)
-          );
-        } else if (headerWriter != null) {
+        if (headerWriter != null) {
+          if (sinkRecord.headers() == null || sinkRecord.headers().isEmpty()) {
+            throw new DataException(
+                String.format("Headers cannot be null for SinkRecord: %s", sinkRecord)
+            );
+          }
           headerWriter.write(sinkRecord);
         }
       }
--- a/kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
+++ b/kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
@@ -21,7 +21,6 @@ import com.amazonaws.services.s3.model.Tag;
 import io.confluent.common.utils.SystemTime;
 import io.confluent.connect.s3.format.KeyValueHeaderRecordWriterProvider;
 import io.confluent.connect.s3.format.RecordViewSetter;
-import io.confluent.connect.s3.format.RecordViews;
 import io.confluent.connect.s3.format.RecordViews.HeaderRecordView;
 import io.confluent.connect.s3.format.RecordViews.KeyRecordView;
 import io.confluent.kafka.serializers.NonRecordContainer;
@@ -76,12 +75,17 @@ import static org.apache.kafka.common.utils.Time.SYSTEM;
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
 public class TopicPartitionWriterTest extends TestWithMockedS3 {
   // The default
   private static final String ZERO_PAD_FMT = "%010d";
-  private enum KVHType { KEYS, HEADERS, VALUES }
+  private enum RecordElement {
+    KEYS,
+    HEADERS,
+    VALUES
+  }
 
   private RecordWriterProvider<S3SinkConnectorConfig> writerProvider;
   private S3Storage storage;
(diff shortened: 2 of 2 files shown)


levzem (commented):
let's add a unit test

dosvath:
Thanks @levzem, I added tests, ready for another pass.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
```diff
@@ -95,7 +95,8 @@ public void write(SinkRecord sinkRecord) {
         }
 
         // headerWriter != null means writing headers is turned on
-        if (headerWriter != null && sinkRecord.headers() == null) {
+        if (headerWriter != null
```
kkonstantine, line 98: @dosvath I find the combination of `if` and then `if else` more confusing than it's required. 

I'd suggest, with the opportunity of this bugfix to change the branches above and here as: 

```
        if (keyWriter != null) {
          if (sinkRecord.key() == null) {
            throw new DataException(
                String.format("Key cannot be null for SinkRecord: %s", sinkRecord)
            );
          }
          keyWriter.write(sinkRecord);
        }
```
and then
```
        if (headerWriter != null) {
          if (sinkRecord.headers() == null || sinkRecord.headers().isEmpty()) {
            throw new DataException(
                String.format("Headers cannot be null for SinkRecord: %s", sinkRecord)
            );
          }
          headerWriter.write(sinkRecord);
        }
```

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -72,6 +81,7 @@
 public class TopicPartitionWriterTest extends TestWithMockedS3 {
   // The default
   private static final String ZERO_PAD_FMT = "%010d";
+  private enum KVHType { KEYS, HEADERS, VALUES }
```
kkonstantine, line 84: I thought we had a checkstyle rule for naming that used consecutive upper case letters. I guess it's not run in tests. 

```suggestion
  private enum RecordElement { 
  KEYS, 
  HEADERS, 
  VALUES 
  }
```
(please fix alignment, github editing is not great for that).

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -900,6 +910,93 @@ public void testAddingS3ObjectTags() throws Exception{
...
+  public void testExceptionOnNullKeys() throws Exception {
+    SinkRecord faultyRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, null,
+        Schema.STRING_SCHEMA, "1", 1, 0L, TimestampType.NO_TIMESTAMP_TYPE, sampleHeaders());
+    writeRecordWithKeysAndHeaders(faultyRecord);
+  }
+
+  @Test (expected = DataException.class)
+  public void testExceptionOnEmptyHeaders() throws Exception {
+    SinkRecord faultyRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, "key",
+        Schema.STRING_SCHEMA, "1", 1, 0L, TimestampType.NO_TIMESTAMP_TYPE, Collections.emptyList());
```
kkonstantine, line 923: What's the difference between "1" and 1?
I'd suggest naming these variables before calling this constructor. 
value = "1" (?)
etc ...

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -900,6 +910,93 @@ public void testAddingS3ObjectTags() throws Exception{
...
   }
 
+  @Test (expected = DataException.class)
+  public void testExceptionOnNullKeys() throws Exception {
+    SinkRecord faultyRecord = new SinkRecord(TOPIC, PARTITION, Schema.STRING_SCHEMA, null,
+        Schema.STRING_SCHEMA, "1", 1, 0L, TimestampType.NO_TIMESTAMP_TYPE, sampleHeaders());
+    writeRecordWithKeysAndHeaders(faultyRecord);
+  }
+
+  @Test (expected = DataException.class)
```
kkonstantine, line 920: Please use `assertThrows`. It's the up-to-date way to assert on exceptions. And even expect other details related to the exception, besides their type.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -900,6 +910,93 @@ public void testAddingS3ObjectTags() throws Exception{
...
+    // setup key record provider for writing record key files.
+    RecordWriterProvider<S3SinkConnectorConfig> keyWriterProvider =
+        new AvroFormat(storage).getRecordWriterProvider();
+    ((RecordViewSetter) keyWriterProvider).setRecordView(new KeyRecordView());
+    // setup header record provider for writing record header files.
+    RecordWriterProvider<S3SinkConnectorConfig> headerWriterProvider =
+        new AvroFormat(storage).getRecordWriterProvider();
+    ((RecordViewSetter) headerWriterProvider).setRecordView(new HeaderRecordView());
+    // initialize the KVHWriterProvider with header and key writers turned on.
+    RecordWriterProvider<S3SinkConnectorConfig> kvhWriterProvider = new KeyValueHeaderRecordWriterProvider(
```
kkonstantine, line 954: or simply, `writerProvider` 
With IDEs these days, there's rarely the need to use acronyms. Speaking as an ex-vim coder :)

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -980,6 +1099,94 @@ private void verify(List<String> expectedFileKeys, int expectedSize, Schema sche
     }
   }
 
+  // based on verify()
+  private void verifyKVH(List<String> expectedFileKeys, int expectedSize, List<SinkRecord> records, KVHType fileType)
```
kkonstantine, line 1103: same as above. Please use a meaningful name.

kkonstantine (commented):
Thanks @dosvath 
Took a pass and left a few comments.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/KeyValueHeaderRecordWriterProvider.java
dosvath: I agree 👍

dosvath:
Thank you for the detailed review @kkonstantine, updated the PR.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -900,6 +914,108 @@ public void testAddingS3ObjectTags() throws Exception{
...
+    expectedHeaderFiles.add(FileUtils.fileKeyToCommit(topicsDir, dirPrefix, TOPIC_PARTITION, 0, ".headers.avro", ZERO_PAD_FMT));
+    expectedHeaderFiles.add(FileUtils.fileKeyToCommit(topicsDir, dirPrefix, TOPIC_PARTITION, 3, ".headers.avro", ZERO_PAD_FMT));
+    expectedHeaderFiles.add(FileUtils.fileKeyToCommit(topicsDir, dirPrefix, TOPIC_PARTITION, 6, ".headers.avro", ZERO_PAD_FMT));
+    verifyRecordElement(expectedHeaderFiles, 3, sinkRecords, RecordElement.HEADERS);
+
+    List<String> expectedKeyFiles = new ArrayList<>();
+    expectedKeyFiles.add(FileUtils.fileKeyToCommit(topicsDir, dirPrefix, TOPIC_PARTITION, 0, ".keys.avro", ZERO_PAD_FMT));
+    expectedKeyFiles.add(FileUtils.fileKeyToCommit(topicsDir, dirPrefix, TOPIC_PARTITION, 3, ".keys.avro", ZERO_PAD_FMT));
+    expectedKeyFiles.add(FileUtils.fileKeyToCommit(topicsDir, dirPrefix, TOPIC_PARTITION, 6, ".keys.avro", ZERO_PAD_FMT));
+    verifyRecordElement(expectedKeyFiles, 3, sinkRecords, RecordElement.KEYS);
```
levzem, line 1016: can put this in one for loop

levzem (approved):
1 nit, LGTM

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
dosvath: If you notice, each block is passing its own different extension; I also think it's clearer to the reader of the test this way the number of files we expect and the extension we are expecting.

PR #378, opened 2020-11-18
CC-12695: stop throwing retriable exceptions on non-retriable S3 errors

levzem:
Signed-off-by: Lev Zemlyanov <lev@confluent.io>

## Problem
the connector will retry non-retriable S3 exceptions i.e `InvalidAccessKey`

if [initiating an upload](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.0.x/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java#L206-L210) throws an exception, it is wrapped in an `IOException`, which follows from [here](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.0.x/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java#L130) to [here](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.0.x/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java#L156)
where it is [swallowed and rethrown](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.0.x/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java#L160-L162) as a `DataException` and [propagated to all of the RecordWriters](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.0.x/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java#L97-L101) which is then propagated to [committing files in TPW](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.0.x/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L519) and [then finally caught and rethrown](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/5.0.x/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L493-L500) as a `RetriableException` because `DataException instanceof ConnectException == true` which results in the faulty retrying of exceptions that should fail the connector

## Solution
do not retry every exception in TPW, instead let the exceptions be propagated as necessary.
to keep existing behavior, the `RetriableExceptions` are now thrown by the `RecordWriters` on failed commits and will not swallow existing `ConnectExceptions`
in `S3OutputStream`, catch and handle the different exceptions properly and propagate non-retriable exceptions as `ConnectExceptions` and all others as `IOExceptions` like previous behavior

bonus: fixed a lot of deprecation warnings

- [ ] yes
- [x] no

## Test Strategy
add tests that make sure the exceptions are properly propagated
fix the old tests to ensure the expected behavior matches

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `5.0.x` where this retrying functionality was added in #281

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
@@ -16,7 +16,9 @@
 
 package io.confluent.connect.s3.storage;
 
-import com.amazonaws.AmazonClientException;
+import com.amazonaws.AmazonServiceException;
+import com.amazonaws.AmazonServiceException.ErrorType;
+import com.amazonaws.SdkClientException;
 import com.amazonaws.event.ProgressEvent;
 import com.amazonaws.event.ProgressListener;
 import com.amazonaws.services.s3.AmazonS3;
@@ -32,6 +34,7 @@ import com.amazonaws.services.s3.model.SSEAwsKeyManagementParams;
 import com.amazonaws.services.s3.model.UploadPartRequest;
 import io.confluent.connect.s3.S3SinkConnectorConfig;
 import io.confluent.connect.storage.common.util.StringUtils;
+import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.errors.DataException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -157,7 +160,7 @@ public class S3OutputStream extends OutputStream {
       }
       multiPartUpload.complete();
       log.debug("Upload complete for bucket '{}' key '{}'", bucket, key);
-    } catch (Exception e) {
+    } catch (IOException e) {
       log.error("Multipart upload failed to complete for bucket '{}' key '{}'", bucket, key);
       throw new DataException("Multipart upload failed to complete.", e);
     } finally {
@@ -204,10 +207,18 @@ public class S3OutputStream extends OutputStream {
 
     try {
       return new MultipartUpload(s3.initiateMultipartUpload(initRequest).getUploadId());
-    } catch (AmazonClientException e) {
-      // TODO: elaborate on the exception interpretation. If this is an AmazonServiceException,
-      // there's more info to be extracted.
-      throw new IOException("Unable to initiate MultipartUpload: " + e, e);
+    } catch (AmazonServiceException e) {
+      if (e.getErrorType() == ErrorType.Client) {
+        // S3 documentation states that this error type means there is a problem with the request
+        // and that retrying this request will not result in a successful response. This includes
+        // errors such as incorrect access keys, invalid parameter values, missing parameters, etc.
+        // Therefore, the connector should propagate this exception and fail.
+        throw new ConnectException("Unable to initiate MultipartUpload", e);
+      }
+
+      throw new IOException("Unable to initiate MultipartUpload.", e);
+    } catch (SdkClientException e) {
+      throw new IOException("Unable to initiate MultipartUpload.", e);
     }
   }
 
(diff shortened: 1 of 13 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -204,10 +207,18 @@ private MultipartUpload newMultipartUpload() throws IOException {
...
-      throw new IOException("Unable to initiate MultipartUpload: " + e, e);
+    } catch (AmazonServiceException e) {
+      if (e.getErrorType() == ErrorType.Client) {
+        // S3 documentation states that this error type means there is a problem with the request
+        // and that retrying this request will not result in a successful response. This includes
+        // errors such as incorrect access keys, invalid parameter values, missing parameters, etc.
+        // Therefore, the connector should propagate this exception and fail.
+        throw new ConnectException("Unable to initiate MultipartUpload", e);
+      }
+
```
kkonstantine, line 218:
```suggestion
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -204,10 +207,18 @@ private MultipartUpload newMultipartUpload() throws IOException {
...
+      throw new IOException("Unable to initiate MultipartUpload.", e);
+    } catch (SdkClientException e) {
```
kkonstantine, line 220: `SdkClientException` is inheriting from `AmazonClientException`. Why not keep the latter?

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterJsonTest.java
```diff
@@ -46,8 +46,8 @@
 import io.confluent.connect.storage.partitioner.Partitioner;
 
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
```
kkonstantine, line 49: please add a comment when the reason of the change is not obvious

kkonstantine (commented):
Looks good overall. Couple of comments.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
levzem: 1. `S3#initiateMultipartUpload` explicitly throws `SDKClientException` and `AmazonServiceException` so I thought it was better to be explicit
2. `AmazonClientException` does not have an `ErrorType` field

kafka-connect-s3/src/test/java/io/confluent/connect/s3/DataWriterJsonTest.java
levzem: mentioned in the PR description, this fixes deprecation warnings

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
kkonstantine: You'll first catch `AmazonServiceException` as you do. 
The follow up will be `AmazonClientException ` for which you don't inspect the error type. 

You can always do `instanceof SdkClientException` in that block if you needed the error type. 
I'm just interested not to risk a gap with this change here. We can revisit exception handling here more holistically in a follow up

kkonstantine (commented):
Thanks @levzem 
Replied inline. 
Also, any idea why the build failed?

kkonstantine (approved):
LGTM!

PR #392, opened 2021-03-02
CCMSG-861: Catch additional exception to report in DLQ

dosvath:
## Problem
There is an additional opportunity to catch and report an exception to the DLQ. 

## Solution
Catch exception on `encodePartition` and report it to the DLQ using the reporter if available. 

- [ ] yes
- [x] no

##### Testing done:
- [X] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
`10.0.x` the earliest DLQ branch.

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -16,4 +16,9 @@
             files="(S3SinkConnectorConfig).java"
     />
 
+    <suppress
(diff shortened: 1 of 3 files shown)


levzem (approved):
LGTM

Issue #411, opened 2021-04-30
Retry AWS server exceptions for MultipartUpload


petoju:
IOException on flushing should be automatically retried.

There is line
https://github.com/confluentinc/kafka-connect-storage-cloud/blob/3e0100422cadacd9aa4f31d3e1a5712005a27880/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java#L232
-> see comment few lines higher, that clarifies, that this should be probably retryable exception. In the end, IOException can happen at any time eg. because of connection issues.

However, the aforementioned exception can bubble up (thru lines 139 and 165) to
https://github.com/confluentinc/kafka-connect-storage-cloud/blob/3e0100422cadacd9aa4f31d3e1a5712005a27880/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java#L171
that is Connect exception and that is not retried.

sejanick:
Hi, we are running into this exact same issue.  It means that transient network/upload failures to AWS are not being appropriately retried in all cases.

robermar23:
We are running into this same issue.  Was wondering what anyone does to work around this as it happens fairly frequently for us.  at least once a day.

chkelly:
Seeing the same issue here as well.

PR #423, opened 2021-05-24
Retry AWS Failures for Multipart Uploads

sejanick:
## Problem
https://github.com/confluentinc/kafka-connect-storage-cloud/issues/411

IOExceptions during multipart uploads should be retried.

We see random failures on multipart uploads fairly regularly, and a single issue puts the connector task in a FAILED state when we do.

## Solution

When an IOException is caught during a multipart upload, also throw an IOException.  The unit test method names (`testPropagateRetriableS3Exceptions()` and  `testPropagateOtherRetriableS3Exceptions()`) already note that this error should be retriable.

- [ ] yes
- [x] no

##### If yes, where?
n/a

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

I would like to merge this change to master.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
@@ -168,7 +168,7 @@ public class S3OutputStream extends PositionOutputStream {
       log.debug("Upload complete for bucket '{}' key '{}'", bucket, key);
     } catch (IOException e) {
       log.error("Multipart upload failed to complete for bucket '{}' key '{}'", bucket, key);
-      throw new ConnectException(
+      throw new IOException(
           String.format("Multipart upload failed to complete: %s", e.getMessage())
       );
     } finally {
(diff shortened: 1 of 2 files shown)


sejanick:
[clabot:check]

sejanick:
[clabot:check]

raphaelauv:
> and a single issue kills the worker immediately when we do

you mean a single issue of multipart set the connect task in a FAILED status, right ? 

( not ALL the task of the connect fail and not the kafka-connect worker itself fail )

sriggin:
> > and a single issue kills the worker immediately when we do
> 
> you mean a single issue of multipart set the connect task in a FAILED status, right ?
> 
> ( not ALL the task of the connect fail and not the kafka-connect worker itself fail )

This is correct. If we have 8 tasks, we can tolerate 8 multipart failures before the entire connector is FAILED, requiring manual intervention.

raphaelauv:
alright , thanks for the precision , could you correct you're pull resquest message

> .. kills the worker 

by

> .. put the connector task in FAILED state

sriggin:
@sejanick is out this week on a vacation, her lack of response is not a lack of interest 😄

sejanick:
@raphaelauv Thanks, I updated the description of this PR with your suggestion!

sejanick:
After more testing and running this particular fix in our production environment, the fix suggested in this PR does not actually solve the problem.  I'm going to close this PR.

We are still trying to figure out a solution to the lack of retries for AWS multipart upload issues.  This is still a real pain point for us.

PR #425, opened 2021-05-25
Fix build failures due to missing dependency

ilanjiR:
## Problem
There was a missing dependency caused by an upstream update to zookeeper

## Solution
Add the missing dependency 

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests


kkonstantine:
Fixed by https://github.com/confluentinc/kafka-connect-storage-common/pull/192 instead

Issue #413, opened 2021-05-05
Exception in `JsonRecordWriterProvider` for `RecordViews.HeaderRecordView`


BabkenSahakyan:
Hi.
I've configured s3 sink connect to push messages with headers to MinIO server.

Here is connector configurations related to headers part:
```
"header.converter": "org.apache.kafka.connect.storage.SimpleHeaderConverter",
"store.kafka.headers": "true",
"headers.format.class": "io.confluent.connect.s3.format.json.JsonFormat"
```

and some example headers copied from control-center UI:
```json
[
  {
    "key": "modifiedBy",
    "stringValue": "BabkenSahakyan"
  },
  {
    "key": "operation",
    "stringValue": "EDIT"
  }
]
```

But, I'm getting `ObjectMapper` serialization exception, saying that it cannot serialize `Struct`, on line [89](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/3e0100422cadacd9aa4f31d3e1a5712005a27880/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java#L89) of `JsonRecordWriterProvider` because return type of `recordView.getView(record, false);` (line [79](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/3e0100422cadacd9aa4f31d3e1a5712005a27880/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java#L79)) is `List<Struct>`, not a `Struct` for `enveloped` argument is false and `RecordViews.HeaderRecordView` is returning [plain](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/3e0100422cadacd9aa4f31d3e1a5712005a27880/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/RecordViews.java#L104) value.

So what I did to fix it:
[...]

PR #431, opened 2021-06-01
CC-14987: Fix serialization when headers.format.class is json

dosvath:
## Problem
Using the configs below results in a serialization exception. 
```
"store.kafka.headers": "true",
"headers.format.class": "io.confluent.connect.s3.format.json.JsonFormat"
```
Error
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: No serializer found for class org.apache.kafka.connect.data.Struct and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) (through reference chain: java.util.ArrayList[0])
	at com.fasterxml.jackson.databind.exc.InvalidDefinitionException.from(InvalidDefinitionException.java:77)
	at com.fasterxml.jackson.databind.SerializerProvider.reportBadDefinition(SerializerProvider.java:1191)
```

Reported in https://github.com/confluentinc/kafka-connect-storage-cloud/issues/413.

**Root Cause** 
The non-empty header array is attempted to be written with the Json writer that expects a different format. 

## Solution

Envelop the header array in a struct so that it can be written without a serialization issue. The data is enveloped by setting the flag `true` in: `recordView.getView(record, true);` and `recordView.getViewSchema(record, true);`

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Backporting to `10.0.x` where header feature was added.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
@@ -76,24 +76,14 @@ public class JsonRecordWriterProvider extends RecordViewSetter
         @Override
         public void write(SinkRecord record) {
           log.trace("Sink record with view {}: {}", recordView, record);
+          // headers need to be enveloped for json format
+          boolean envelop = recordView instanceof HeaderRecordView;
           try {
-            // headers need to be enveloped for json format
-            if (recordView instanceof HeaderRecordView) {
-              byte[] rawJson = converter.fromConnectData(
-                  record.topic(),
-                  recordView.getViewSchema(record, true),
-                  recordView.getView(record, true)
-              );
-              s3outWrapper.write(rawJson);
-              s3outWrapper.write(LINE_SEPARATOR_BYTES);
-              return;
-            }
-
-            Object value = recordView.getView(record, false);
+            Object value = recordView.getView(record, envelop);
             if (value instanceof Struct) {
               byte[] rawJson = converter.fromConnectData(
                   record.topic(),
-                  recordView.getViewSchema(record, false),
+                  recordView.getViewSchema(record, envelop),
                   value
               );
               s3outWrapper.write(rawJson);
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
@@ -19,6 +19,7 @@ import static io.confluent.connect.s3.util.Utils.getAdjustedFilename;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.confluent.connect.s3.format.RecordViews.HeaderRecordView;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.errors.RetriableException;
@@ -76,6 +77,18 @@ public class JsonRecordWriterProvider extends RecordViewSetter
         public void write(SinkRecord record) {
           log.trace("Sink record with view {}: {}", recordView, record);
           try {
+            // headers need to be enveloped for json format
+            if (recordView instanceof HeaderRecordView) {
+              byte[] rawJson = converter.fromConnectData(
+                  record.topic(),
+                  recordView.getViewSchema(record, true),
+                  recordView.getView(record, true)
+              );
+              s3outWrapper.write(rawJson);
+              s3outWrapper.write(LINE_SEPARATOR_BYTES);
+              return;
+            }
+
             Object value = recordView.getView(record, false);
             if (value instanceof Struct) {
               byte[] rawJson = converter.fromConnectData(
--- a/kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
+++ b/kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
@@ -29,6 +29,8 @@ import io.confluent.connect.s3.format.json.JsonFormat;
 import io.confluent.connect.s3.storage.CompressionType;
 import io.confluent.connect.storage.errors.PartitionException;
 import io.confluent.kafka.serializers.NonRecordContainer;
+import java.util.HashSet;
+import java.util.Set;
 import org.apache.avro.util.Utf8;
 import org.apache.kafka.common.record.TimestampType;
 import org.apache.kafka.connect.connector.ConnectRecord;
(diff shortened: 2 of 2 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
```diff
@@ -76,6 +77,18 @@ public RecordWriter getRecordWriter(final S3SinkConnectorConfig conf, final Stri
         public void write(SinkRecord record) {
           log.trace("Sink record with view {}: {}", recordView, record);
           try {
+            // headers need to be enveloped for json format
```
levzem, line 80: does this envelope still work for avro and other formats?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
```diff
@@ -76,6 +77,18 @@ public RecordWriter getRecordWriter(final S3SinkConnectorConfig conf, final Stri
...
+                  record.topic(),
+                  recordView.getViewSchema(record, true),
+                  recordView.getView(record, true)
+              );
+              s3outWrapper.write(rawJson);
+              s3outWrapper.write(LINE_SEPARATOR_BYTES);
+              return;
+            }
+
             Object value = recordView.getView(record, false);
```
levzem, line 92: boolean envelope = recordView instanceof HeaderRecordView;
```suggestion
            Object value = recordView.getView(record, envelope);
```

why can't we do something like this?

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -1312,6 +1372,12 @@ private void verifyRecordElement(List<String> expectedFileKeys, int expectedSize
           Schema headerSchema = new HeaderRecordView().getViewSchema(currentRecord, false);
           Object value = new HeaderRecordView().getView(currentRecord, false);
           expectedRecord = ((NonRecordContainer) format.getAvroData().fromConnectData(headerSchema, value)).getValue();
+        } else if (fileKey.endsWith(".headers.json")) {
```
levzem, line 1375: static variable

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -1335,7 +1413,7 @@ private boolean filenameContainsExtensions(String filename, List<String> extensi
 
   // filter for values only.
   private List<String> getS3FileListValues(List<S3ObjectSummary> summaries) {
-    List<String> excludeExtensions = Arrays.asList(".headers.avro", ".keys.avro");
+    List<String> excludeExtensions = Arrays.asList(".headers.avro", ".headers.json", ".keys.avro");
```
levzem, line 1416:
```suggestion
    Set<String> excludeExtensions = Arrays.asList(".headers.avro", ".headers.json", ".keys.avro");
```

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -1176,6 +1218,19 @@ public void testRecordKeysAndHeadersWritten() throws Exception {
...
   }
 
+  private RecordWriterProvider<S3SinkConnectorConfig> getKeyHeaderValueProviderJsonHeaders() {
+    // setup header record provider for writing record header files.
+    RecordWriterProvider<S3SinkConnectorConfig> headerWriterProvider =
+        new JsonFormat(storage).getRecordWriterProvider();
+    ((RecordViewSetter) headerWriterProvider).setRecordView(new HeaderRecordView());
+    // initialize the KVHWriterProvider with header and key writers turned on.
+    return new KeyValueHeaderRecordWriterProvider(
+        new AvroFormat(storage).getRecordWriterProvider(),
```
levzem, line 1228: why is this avro?

levzem (commented):
overall lgtm

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
dosvath: Headers don't require the additional enveloped format when using the other writer classes. It's a unique requirement for headers with the Json writer.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
dosvath: Sounds good.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
dosvath: The first parameter for the `KeyValueHeaderRecordWriterProvider()` refers to the converter for the record value, for the tests in this PR the main focus is the converter for the headers to be json, the first converter could be any of the formats available.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
levzem: right, so we are enveloping all of the headers regardless of the format now, is that correct? while before we did not envelop avro, which means this fix isn't backwards incompatible or am i missing something?

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -1079,6 +1090,46 @@ public void testRecordKeysAndHeadersWritten() throws Exception {
     verifyRecordElement(expectedKeyFiles, 3, sinkRecords, RecordElement.KEYS);
   }
 
+  @Test
+  public void testRecordHeadersWrittenJson() throws Exception {
```
levzem, line 1094: tests look good, but looks like there is a good opportunity for parametrizing these tests across all of the formats, might be worth looking into but not necessary for this PR

levzem (commented):
one last question above otherwise LGTM

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
dosvath: This change is only for the `JsonRecordWriterProvider`. `Avro` and other formats are not affected by this PR.

levzem: I am blind and I apologize

levzem (approved):
LGTM

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonRecordWriterProvider.java
dosvath: all good, complex feature :)

PR #432, opened 2021-06-03
MINOR: Remove unused dependency on kafka-connect-storage-hive in the 5.2.x release branch

kkonstantine:
## Problem
This package is not used but bring in several transitive deps

## Solution
We should remove and align with more recent branches. 

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
Applies only to 5.2.x

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,10 +108,6 @@
         </dependency>
         <dependency>
             <groupId>io.confluent</groupId>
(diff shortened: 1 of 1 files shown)


levzem (approved):
LGTM

PR #442, opened 2021-07-29
CCMSG-1194: Upgrade parent and storage-common dependencies to 10.2.1

kkonstantine:
## Problem
Parent has fixes that now need to be ported with an explicit bump

## Solution
Bump the parent to the next patch release

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


dosvath (approved):
LGTM pending green build

Issue #235, opened 2019-03-19
S3 connector - intermittent no route to host - does not retry


csghuser:
We have kafka-connect configured with a number of topics, all writing out to s3. 

The version of the s3 connector we are using is:

`confluent-kafka-connect-s3-4.1.2-1.noarch`

The connector is running outside of AWS. The following error was seen:

```
[2019-03-19 01:01:49,189] INFO Files committed to S3. Target commit offset for some-topic-0 is 28425321 (io.confluent.connect.s3.TopicPartitionWriter:491)
[2019-03-19 01:01:55,466] ERROR Multipart upload failed to complete for bucket 'some-aws-bucket' key 'some-dest/somefile..json' (io.confluent.connect.s3.storage.S3OutputStream:162)
[2019-03-19 01:01:55,466] ERROR WorkerSinkTask{id=some-topic-0} Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted. (org.apache.kafka.connect.runtime.WorkerSinkTask:544)
org.apache.kafka.connect.errors.DataException: Multipart upload failed to complete.
        at io.confluent.connect.s3.storage.S3OutputStream.commit(S3OutputStream.java:163)
        at io.confluent.connect.s3.format.json.JsonRecordWriterProvider$1.commit(JsonRecordWriterProvider.java:98)
        at io.confluent.connect.s3.TopicPartitionWriter.commitFile(TopicPartitionWriter.java:503)
        at io.confluent.connect.s3.TopicPartitionWriter.commitFiles(TopicPartitionWriter.java:483)
        at io.confluent.connect.s3.TopicPartitionWriter.executeState(TopicPartitionWriter.java:223)
        at io.confluent.connect.s3.TopicPartitionWriter.write(TopicPartitionWriter.java:176)
        at io.confluent.connect.s3.S3SinkTask.put(S3SinkTask.java:195)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:524)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:302)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:205)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:173)
[...]

PR #461, opened 2021-11-10
Fix retries for S3 transient errors

lion-simba:
## Problem
#411 
#235 

Currently, there are two retry mechanisms supported by Amazon S3 Sink Connector:

1. One is handled by AWS SDK inside of the connector and controlled by `s3.part.retries` setting.
2. The other one is handled by Kafka Connect framework and controlled by `RetriableException`.

The first one works fine. However, there seem to be no ideal value for `s3.part.retries` property:
If it's set too low, in case if S3 continue generating errors, Connector task dies and requires **manual restart**.
If it's set too high, a processing thread gets blocked for a long time. If this time exceeds kafka's `max.poll.interval`, the task get removed from the consumer group, triggering a rebalance. The worst part is (for Kafka versions prior to 2.5) that rebalance **prevents** other (alive) **tasks** of the same consumer group **from processing data** for up to another `max.poll.interval`.

The second retry mechanism doesn't work for Amazon S3 Sink Connector, because in most places in the code IOException is treated as fatal and is re-thrown as non-retriable ConnectException.

If we make second mechanism work, it will be safe to set `s3.part.retries` to rather low values, without the fear that all tasks will go to FAILED state. And in the same time, we would not get unnecessary Kafka rebalances.

## Solution

Main solution is to treat IOException as RetriableException.

Also, `TopicPartitionWriter::currentStartOffset()` is now calculated on demand rather than in the beginning of `commit()` method. Otherwise, if IOException happens during `partUpload()` but before first call to `commit()`, Connector doesn't rewind kafka offset back in `S3SinkTask.java:239`, which leads to a data loss.

- [ ] yes
- [x] no

##### If yes, where?
n/a

##### Testing done:
- [ ] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

Most of the code changes in this PR are for tests.

`TestWithMockedFaultyS3` introduces WireMock proxy, which stands between Amazon S3 Sink Connector and mocked S3 instance provided by `TestWithMockedS3`. WireMock proxy is able to inject errors in the communication channel between Connector and S3 mock.

`S3SinkConnectorFaultyS3Test` is a parametrized test based on `TestWithMockedFaultyS3`. In total, it contains 72 tests. These are basically all combinations of the following dimensions:

1. Output format: `ByteArrayFormat`, `JsonFormat`, `AvroFormat`, `ParquetFormat`. These all are needed because changes were made to each of the corresponding `*WriterProvider` classes.
2. Flush size: `FLUSH_SIZE_SMALL`, `FLUSH_SIZE_BIG`. Depending on how `flush.size` and `s3.part.size` relate to each other, different code path is executed (calling `uploadPart()` from `S3OutputStream::commit()` method vs calling `uploadPart()` from `S3OutputStream::write()` methods).
3. S3 method (phase of a file upload): CreateMultipartUpload, UploadPart, CompleteMultipartUpload
4. Errors: HTTP 429 (Too Many Requests), HTTP 500 (Server Error), Connection Reset

In each test, a new instance of Amazon S3 Sink Connector is configured, failure is injected, and then S3 bucket is inspected for the files.

On my machine, most tests are completed within 2 seconds. However, tests for `ParquetFormat` take up to 40 seconds to complete. This is because `ParquetFormat` has internal buffer set to 128 MB by default. So, to trigger flushing of that buffer to stream, we need to actually write (and generate) this data. Generating and writing such a big amount of (random) data takes time.

## Release Plan
I would like to merge this change to master.
If it's successful, then would like to back port to version 5.5.x (as it is planned to deploy to our production soon).

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -190,6 +190,8 @@ public class TopicPartitionWriter {
       failureTime = -1;
     }
 
+    resetExpiredScheduledRotationIfNoPendingRecords(now);
+
     while (!buffer.isEmpty()) {
       try {
         executeState(now);
@@ -337,6 +339,12 @@ public class TopicPartitionWriter {
     }
   }
 
+  private void resetExpiredScheduledRotationIfNoPendingRecords(long now) {
+    if (recordCount == 0 && shouldApplyScheduledRotation(now)) {
+      setNextScheduledRotation();
+    }
+  }
+
   public void close() throws ConnectException {
     log.debug("Closing TopicPartitionWriter {}", tp);
     for (RecordWriter writer : writers.values()) {
@@ -395,14 +403,18 @@ public class TopicPartitionWriter {
     );
 
     log.trace(
-        "Checking rotation on time with recordCount '{}' and encodedPartition '{}'",
+        "Checking rotation on time for topic-partition '{}' "
+            + "with recordCount '{}' and encodedPartition '{}'",
+        tp,
         recordCount,
         encodedPartition
     );
 
     log.trace(
-        "Should apply periodic time-based rotation (rotateIntervalMs: '{}', baseRecordTimestamp: "
+        "Should apply periodic time-based rotation for topic-partition '{}':"
+            + " (rotateIntervalMs: '{}', baseRecordTimestamp: "
             + "'{}', timestamp: '{}', encodedPartition: '{}', currentEncodedPartition: '{}')? {}",
+        tp,
         rotateIntervalMs,
         baseRecordTimestamp,
         recordTimestamp,
@@ -410,17 +422,22 @@ public class TopicPartitionWriter {
         currentEncodedPartition,
         periodicRotation
     );
+    return periodicRotation || shouldApplyScheduledRotation(now);
+  }
 
+  private boolean shouldApplyScheduledRotation(long now) {
     boolean scheduledRotation = rotateScheduleIntervalMs > 0 && now >= nextScheduledRotation;
     log.trace(
-        "Should apply scheduled rotation: (rotateScheduleIntervalMs: '{}', nextScheduledRotation:"
+        "Should apply scheduled rotation for topic-partition '{}':"
+            + " (rotateScheduleIntervalMs: '{}', nextScheduledRotation:"
             + " '{}', now: '{}')? {}",
+        tp,
         rotateScheduleIntervalMs,
         nextScheduledRotation,
         now,
         scheduledRotation
     );
-    return periodicRotation || scheduledRotation;
+    return scheduledRotation;
   }
 
   private void setNextScheduledRotation() {
@@ -433,8 +450,13 @@ public class TopicPartitionWriter {
       );
       if (log.isDebugEnabled()) {
         log.debug(
-            "Update scheduled rotation timer. Next rotation for {} will be at {}",
+            "Update scheduled rotation timer for topic-partition '{}': "
+                + "(rotateScheduleIntervalMs: '{}', nextScheduledRotation: '{}', now: '{}'). "
+                + "Next rotation will be at {}",
             tp,
+            rotateScheduleIntervalMs,
+            nextScheduledRotation,
+            now,
             new DateTime(nextScheduledRotation).withZone(timeZone).toString()
         );
       }
@@ -443,8 +465,9 @@ public class TopicPartitionWriter {
 
   private boolean rotateOnSize() {
     boolean messageSizeRotation = recordCount >= flushSize;
-    log.trace(
-        "Should apply size-based rotation (count {} >= flush size {})? {}",
+    log.trace("Should apply size-based rotation for topic-partition '{}':"
+            + " (count {} >= flush size {})? {}",
+        tp,
         recordCount,
         flushSize,
         messageSizeRotation
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -74,7 +74,6 @@ public class TopicPartitionWriter {
   private final long rotateScheduleIntervalMs;
   private long nextScheduledRotation;
   private long currentOffset;
-  private Long currentStartOffset;
   private Long currentTimestamp;
   private String currentEncodedPartition;
   private Long baseRecordTimestamp;
@@ -358,7 +357,7 @@ public class TopicPartitionWriter {
   }
 
   public Long currentStartOffset() {
-    return currentStartOffset;
+    return minStartOffset();
   }
 
   public void failureTime(long when) {
@@ -573,7 +572,6 @@ public class TopicPartitionWriter {
   }
 
   private void commitFiles() {
-    currentStartOffset = minStartOffset();
     for (Map.Entry<String, String> entry : commitFiles.entrySet()) {
       String encodedPartition = entry.getKey();
       commitFile(encodedPartition);
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -76,7 +76,7 @@ public class AvroRecordWriterProvider extends RecordViewSetter
             writer.setCodec(CodecFactory.fromString(conf.getAvroCodec()));
             writer.create(avroSchema, s3out);
           } catch (IOException e) {
-            throw new ConnectException(e);
+            throw new RetriableException(e);
           }
         }
         log.trace("Sink record with view {}: {}", recordView, record);
@@ -89,7 +89,7 @@ public class AvroRecordWriterProvider extends RecordViewSetter
           }
           writer.append(value);
         } catch (IOException e) {
-          throw new ConnectException(e);
+          throw new RetriableException(e);
         }
       }
 
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -17,12 +17,12 @@ package io.confluent.connect.s3.format.avro;
 
 import static io.confluent.connect.s3.util.Utils.getAdjustedFilename;
 
+import io.confluent.connect.s3.util.S3ErrorUtils;
 import org.apache.avro.file.CodecFactory;
 import org.apache.avro.file.DataFileWriter;
 import org.apache.avro.generic.GenericDatumWriter;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.errors.ConnectException;
-import org.apache.kafka.connect.errors.RetriableException;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -76,7 +76,7 @@ public class AvroRecordWriterProvider extends RecordViewSetter
             writer.setCodec(CodecFactory.fromString(conf.getAvroCodec()));
             writer.create(avroSchema, s3out);
           } catch (IOException e) {
-            throw new RetriableException(e);
+            throw S3ErrorUtils.maybeRetriableConnectException(e);
           }
         }
         log.trace("Sink record with view {}: {}", recordView, record);
@@ -89,7 +89,7 @@ public class AvroRecordWriterProvider extends RecordViewSetter
           }
           writer.append(value);
         } catch (IOException e) {
-          throw new RetriableException(e);
+          throw S3ErrorUtils.maybeRetriableConnectException(e);
         }
       }
 
@@ -102,7 +102,7 @@ public class AvroRecordWriterProvider extends RecordViewSetter
           s3out.commit();
           writer.close();
         } catch (IOException e) {
-          throw new RetriableException(e);
+          throw S3ErrorUtils.maybeRetriableConnectException(e);
         }
       }
 
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -15,9 +15,9 @@
 
 package io.confluent.connect.s3.format.avro;
 
+import static io.confluent.connect.s3.util.S3ErrorUtils.throwConnectException;
 import static io.confluent.connect.s3.util.Utils.getAdjustedFilename;
 
-import io.confluent.connect.s3.util.S3ErrorUtils;
 import org.apache.avro.file.CodecFactory;
 import org.apache.avro.file.DataFileWriter;
 import org.apache.avro.generic.GenericDatumWriter;
@@ -76,7 +76,7 @@ public class AvroRecordWriterProvider extends RecordViewSetter
             writer.setCodec(CodecFactory.fromString(conf.getAvroCodec()));
             writer.create(avroSchema, s3out);
           } catch (IOException e) {
-            throw S3ErrorUtils.maybeRetriableConnectException(e);
+            throwConnectException(e);
           }
         }
         log.trace("Sink record with view {}: {}", recordView, record);
@@ -89,7 +89,7 @@ public class AvroRecordWriterProvider extends RecordViewSetter
           }
           writer.append(value);
         } catch (IOException e) {
-          throw S3ErrorUtils.maybeRetriableConnectException(e);
+          throwConnectException(e);
         }
       }
 
@@ -102,7 +102,7 @@ public class AvroRecordWriterProvider extends RecordViewSetter
           s3out.commit();
           writer.close();
         } catch (IOException e) {
-          throw S3ErrorUtils.maybeRetriableConnectException(e);
+          throwConnectException(e);
         }
       }
 
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
@@ -34,6 +34,7 @@ import com.amazonaws.services.s3.model.UploadPartRequest;
 import io.confluent.connect.s3.S3SinkConnectorConfig;
 import io.confluent.connect.storage.common.util.StringUtils;
 import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.errors.RetriableException;
 import org.apache.parquet.io.PositionOutputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -168,8 +169,8 @@ public class S3OutputStream extends PositionOutputStream {
       log.debug("Upload complete for bucket '{}' key '{}'", bucket, key);
     } catch (IOException e) {
       log.error("Multipart upload failed to complete for bucket '{}' key '{}'", bucket, key);
-      throw new ConnectException(
-          String.format("Multipart upload failed to complete: %s", e.getMessage())
+      throw new RetriableException(
+          String.format("Multipart upload failed to complete: %s", e.getMessage()), e
       );
     } finally {
       buffer.clear();
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
@@ -45,6 +45,7 @@ import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Supplier;
 
 /**
  * Output stream enabling multi-part uploads of Kafka records.
@@ -220,14 +221,26 @@ public class S3OutputStream extends PositionOutputStream {
       initRequest.setSSECustomerKey(sseCustomerKey);
     }
 
+    return handleAmazonExceptions(
+      () -> new MultipartUpload(s3.initiateMultipartUpload(initRequest).getUploadId())
+    );
+  }
+
+  private <T> T handleAmazonExceptions(Supplier<T> supplier) throws IOException {
     try {
-      return new MultipartUpload(s3.initiateMultipartUpload(initRequest).getUploadId());
+      return supplier.get();
     } catch (AmazonServiceException e) {
       if (e.getErrorType() == ErrorType.Client) {
         // S3 documentation states that this error type means there is a problem with the request
         // and that retrying this request will not result in a successful response. This includes
         // errors such as incorrect access keys, invalid parameter values, missing parameters, etc.
         // Therefore, the connector should propagate this exception and fail.
+
+        // The only exception is "Too Many Requests" - these may succeed after some backoff
+        if (e.getStatusCode() == 429) {
+          throw new IOException("Unable to initiate MultipartUpload.", e);
+        }
+
         throw new ConnectException("Unable to initiate MultipartUpload", e);
       }
       throw new IOException("Unable to initiate MultipartUpload.", e);
@@ -266,11 +279,14 @@ public class S3OutputStream extends PositionOutputStream {
       partETags.add(s3.uploadPart(request).getPartETag());
     }
 
-    public void complete() {
+    public void complete() throws IOException {
       log.debug("Completing multi-part upload for key '{}', id '{}'", key, uploadId);
       CompleteMultipartUploadRequest completeRequest =
           new CompleteMultipartUploadRequest(bucket, key, uploadId, partETags);
-      s3.completeMultipartUpload(completeRequest);
+
+      handleAmazonExceptions(
+          () -> s3.completeMultipartUpload(completeRequest)
+      );
     }
 
     public void abort() {
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
@@ -16,8 +16,6 @@
 package io.confluent.connect.s3.storage;
 
 import com.amazonaws.AmazonClientException;
-import com.amazonaws.AmazonServiceException;
-import com.amazonaws.AmazonServiceException.ErrorType;
 import com.amazonaws.event.ProgressEvent;
 import com.amazonaws.event.ProgressListener;
 import com.amazonaws.services.s3.AmazonS3;
@@ -32,9 +30,8 @@ import com.amazonaws.services.s3.model.SSEAwsKeyManagementParams;
 import com.amazonaws.services.s3.model.SSECustomerKey;
 import com.amazonaws.services.s3.model.UploadPartRequest;
 import io.confluent.connect.s3.S3SinkConnectorConfig;
+import io.confluent.connect.s3.util.S3ErrorUtils;
 import io.confluent.connect.storage.common.util.StringUtils;
-import org.apache.kafka.connect.errors.ConnectException;
-import org.apache.kafka.connect.errors.RetriableException;
 import org.apache.parquet.io.PositionOutputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -170,8 +167,9 @@ public class S3OutputStream extends PositionOutputStream {
       log.debug("Upload complete for bucket '{}' key '{}'", bucket, key);
     } catch (IOException e) {
       log.error("Multipart upload failed to complete for bucket '{}' key '{}'", bucket, key);
-      throw new RetriableException(
-          String.format("Multipart upload failed to complete: %s", e.getMessage()), e
+      throw S3ErrorUtils.maybeRetriableConnectException(
+          String.format("Multipart upload failed to complete: %s", e.getMessage()),
+          e
       );
     } finally {
       buffer.clear();
@@ -226,24 +224,19 @@ public class S3OutputStream extends PositionOutputStream {
     );
   }
 
-  private <T> T handleAmazonExceptions(Supplier<T> supplier) throws IOException {
+  /**
+   * Return the given Supplier value, converting any thrown AmazonClientException object
+   * to an IOException object (containing the AmazonClientException object) and
+   * throw that instead.
+   * @param supplier The supplier to evaluate
+   * @param <T> The object type returned by the Supplier
+   * @return The value returned by the Supplier
+   * @throws IOException Any IOException or AmazonClientException thrown while
+   *                     retreiving the Supplier's value
+   */
+  private static <T> T handleAmazonExceptions(Supplier<T> supplier) throws IOException {
     try {
       return supplier.get();
-    } catch (AmazonServiceException e) {
-      if (e.getErrorType() == ErrorType.Client) {
-        // S3 documentation states that this error type means there is a problem with the request
-        // and that retrying this request will not result in a successful response. This includes
-        // errors such as incorrect access keys, invalid parameter values, missing parameters, etc.
-        // Therefore, the connector should propagate this exception and fail.
-
-        // The only exception is "Too Many Requests" - these may succeed after some backoff
-        if (e.getStatusCode() == 429) {
-          throw new IOException(e);
-        }
-
-        throw new ConnectException(e);
-      }
-      throw new IOException(e);
     } catch (AmazonClientException e) {
       throw new IOException(e);
     }
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
@@ -30,7 +30,6 @@ import com.amazonaws.services.s3.model.SSEAwsKeyManagementParams;
 import com.amazonaws.services.s3.model.SSECustomerKey;
 import com.amazonaws.services.s3.model.UploadPartRequest;
 import io.confluent.connect.s3.S3SinkConnectorConfig;
-import io.confluent.connect.s3.util.S3ErrorUtils;
 import io.confluent.connect.storage.common.util.StringUtils;
 import org.apache.parquet.io.PositionOutputStream;
 import org.slf4j.Logger;
@@ -52,7 +51,6 @@ import java.util.function.Supplier;
 public class S3OutputStream extends PositionOutputStream {
   private static final Logger log = LoggerFactory.getLogger(S3OutputStream.class);
   private final AmazonS3 s3;
-  private final S3SinkConnectorConfig connectorConfig;
   private final String bucket;
   private final String key;
   private final String ssea;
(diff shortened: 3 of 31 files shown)


kpatelatwork:
Hey Lion let me put this in pipeline for roadmap

lion-simba:
Hi @kpatelatwork, thanks for looking into this!

Is there any date (approximate?) I can expect for this PR to move forward?

I'm especially interested in code review approval (or feedback) from someone of Confluent experts. This will add me additional confidence that the change is safe and I didn't miss anything, so that we can move forward and deploy it in our company without waiting for the actual release from Confluent.

cjolivier01:
> Hi @kpatelatwork, thanks for looking into this!
> 
> Is there any date (approximate?) I can expect for this PR to move forward?
> 
> I'm especially interested in code review approval (or feedback) from someone of Confluent experts. This will add me additional confidence that the change is safe and I didn't miss anything, so that we can move forward and deploy it in our company without waiting for the actual release from Confluent.

Hi @lion-simba This change looks reasonable.  I have scheduled the review and testing of this for this coming sprint (since it has a significant blast radius), so it may be a couple of weeks before it will be merged in.  Is this acceptable?

kpatelatwork:
Hey @cjolivier01 we have Thanksgiving coming in 2 weeks so I think we should plan a buffer and lets talk internally in sprint planning next week and then we can give some guidance to @lion-simba .

lion-simba:
Hi @kpatelatwork, @cjolivier01,

> so it may be a couple of weeks before it will be merged in. Is this acceptable?

> we should plan a buffer and lets talk internally in sprint planning next week

Sure, it's acceptable. I mean - plan as you need. Just let me know if anything changes. Also, let me know if you need any help with this from me.

I appreciate you are working with this PR.

lion-simba:
Hi @kpatelatwork, @cjolivier01,

Any updates on this PR?

kpatelatwork:
Hey Alexey unfortunately no updates so far but we will keep you posted when
its added in sprint.

raphaelauv:
Thanks you @lion-simba !

The data loss `minStartOffset` could be in a separate PR , I would for sure cherry-pick it to put it in production :+1:

lion-simba:
Hi @raphaelauv,

> The data loss minStartOffset could be in a separate PR , I would for sure cherry-pick it to put it in production +1

It doesn't make sense to extract it into separate PR, because data loss can only happen if RecordWriter::write() method throws RetriableException, which is not the case for the original code (master branch). It's this PR, which changes behavior so that RecordWriter::write() method throws RetriableException instead of ConnectException.

In the original code, the same case would lead to Task being killed. After (manual) restart, a Task would process data from the last committed offset, so no data loss would happen.

lion-simba:
Hi @kpatelatwork,

I noticed that a new Check has been added recently ("Jenkins-public-CI"), and it's failing. I was hoping to find some details in a Jenkins build log, but it looks like "https://jenkins.public.confluent.io" is inaccessible from the Internet. How can I figure out why the check is failing?

kpatelatwork:
Hey @lion-simba I think @cjolivier01 is also running into this on his PR, we will figure this out and let you know.

lion-simba:
Hi @kpatelatwork and @cjolivier01,

"https://jenkins.public.confluent.io" is now accessible from the Internet, and I was able to see a reason why the [check is failing](https://jenkins.public.confluent.io/job/kafka-connect-storage-cloud/job/PR-461/4/). The reason is integration tests (`io.confluent.connect.s3.integration.S3SinkConnectorIT.io.confluent.connect.s3.integration.S3SinkConnectorIT`) have insufficient permissions to write to S3 bucket:

```
com.amazonaws.services.s3.model.AmazonS3Exception: Access Denied (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: 5ZWB2DRWM4YRKQK7; S3 Extended Request ID: FQ1mCV3IJk1hEt99WZi3N2F5C9K5DWp++u7M7jW253tNXIe2eUecvoJYij/4QNHg0ZAjqqNjaY0=; Proxy: null)
	at io.confluent.connect.s3.integration.S3SinkConnectorIT.setupClient(S3SinkConnectorIT.java:156)
```

This failure has nothing to do with this PR and seems like a misconfiguration of build environment.

Is it a blocker for a merge? If yes, when it's planned to be fixed?

PS. I updated the PR with a fresh master and resolved conflicts.

cjolivier01:
@lion-simba Hi Alex, sorry for the delay -- looking at this today.

cjolivier01:
@lion-simba By default, the maximum backoff time is 24 hrs (which I assume is what is causing the max.poll.interval to be exceeded and thus causing a rebalance):

https://github.com/confluentinc/kafka-connect-storage-cloud/blob/master/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java#L160

This value is passed to the AWS backoff strategy here (as you may know): 

https://github.com/confluentinc/kafka-connect-storage-cloud/blob/master/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java#L164

If 
```java
S3_RETRY_MAX_BACKOFF_TIME_MS
```
were made into a configurable item (which could manually, or automatically, coincide with max.poll.interval x number of retries (taking into account exponential back-off as well))  would this also serve the purpose?

A major advantage of this mechanism is that the AWS SDK already manages a list of retry-able exceptions (rather than all IOExceptions thrown).

cjolivier01:
> @lion-simba By default, the maximum backoff time is 24 hrs (which I assume is what is causing the max.poll.interval to be exceeded and thus causing a rebalance):
> 
> ...

@lion-simba After reviewing/testing some more, this looks really great!  I think the above question I posed is somewhat rhetorical now,  However, I would be interested in your input wrt the AWS retry timing portion I mentioned (although that does not cover the desired poll-retry in this PR and some of the other workflows) and how it relates to poll timeout.  I agree that the poll retry is necessary for most of these cases (beyond the AWS-retry).

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -72,23 +64,6 @@ protected void startConnect() {
...
-  /**
-   * Wait up to {@link #CONNECTOR_STARTUP_DURATION_MS maximum time limit} for the connector with the
-   * given name to start the specified number of tasks.
-   *
-   * @param name     the name of the connector
-   * @param numTasks the minimum number of tasks that are expected
-   * @return the time this method discovered the connector has started, in milliseconds past epoch
-   * @throws InterruptedException if this was interrupted
-   */
-  protected long waitForConnectorToStart(String name, int numTasks) throws InterruptedException {
```
cjolivier01, line 84: note: Moved to EmbeddedConnectUtils.java

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -100,74 +75,7 @@ protected long waitForConnectorToStart(String name, int numTasks) throws Interru
...
-  }
-
-  /**
-   * Confirm that a connector with an exact number of tasks is running.
-   *
-   * @param connectorName the connector
-   * @param numTasks      the minimum number of tasks
-   * @return true if the connector and tasks are in RUNNING state; false otherwise
-   */
-  protected Optional<Boolean> assertConnectorAndTasksRunning(String connectorName, int numTasks) {
```
cjolivier01, line 118: note: Moved to EmbeddedConnectUtils.java

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -100,74 +75,7 @@ protected long waitForConnectorToStart(String name, int numTasks) throws Interru
...
-    }
-  }
-
-  /**
-   * Recursively query the bucket to get the total number of files that exist in the bucket.
-   *
-   * @param bucketName the name of the bucket containing the files.
-   * @return the number of files in the bucket
-   */
-  private int getBucketFileCount(String bucketName) {
```
cjolivier01, line 155: note: Moved to S3utils.java

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -100,74 +75,7 @@ protected long waitForConnectorToStart(String name, int numTasks) throws Interru
...
-  }
-
-  /**
-   * Confirm that the file count in a bucket matches the expected number of files.
-   *
-   * @param bucketName the name of the bucket containing the files
-   * @param expectedNumFiles the number of files expected
-   * @return true if the number of files in the bucket match the expected number; false otherwise
-   */
-  protected Optional<Boolean> assertFileCountInBucket(String bucketName, int expectedNumFiles) {
```
cjolivier01, line 140: note: Moved to S3utils.java

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorNetworkIT.java
```diff
@@ -50,12 +50,7 @@ protected long waitForFilesInBucket(
       String bucketName,
       int numFiles)
       throws InterruptedException {
-    TestUtils.waitForCondition(
-        () -> assertFileCountInBucket(bucketName, numFiles).orElse(false),
```
cjolivier01, line 54: note: logic moved to S3Utils. waitForFilesInBucket()

kafka-connect-s3/src/test/java/io/confluent/connect/s3/util/EmbeddedConnectUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.apache.kafka.connect.runtime.AbstractStatus;
+import org.apache.kafka.connect.runtime.rest.entities.ConnectorStateInfo;
+import org.apache.kafka.connect.util.clusters.EmbeddedConnectCluster;
+import org.apache.kafka.test.TestUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
```
cjolivier01, line 11: note: contains reused functions moved from elsewhere

kafka-connect-s3/src/test/java/io/confluent/connect/s3/util/S3Utils.java
```diff
@@ -0,0 +1,74 @@
...
+import com.amazonaws.services.s3.model.ListObjectsV2Request;
+import com.amazonaws.services.s3.model.ListObjectsV2Result;
+import org.apache.kafka.test.TestUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+
+public class S3Utils {
+    private static final Logger log = LoggerFactory.getLogger(S3Utils.class);
```
cjolivier01, line 13: note: contains reused functions moved from elsewhere

cjolivier01 (commented):
Marked up some comments for people familiar with the current codebase so that the PR doesn;t appear to be more large/complex than it actually is.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/S3SinkConnectorFaultyS3Test.java
```diff
@@ -0,0 +1,389 @@
...
+
+    protected static final String CONNECTOR_NAME = "s3-sink-";
+    protected static final long S3_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(60);
+
+    protected static final String[] TEST_MESSAGES = generateTestMessages(70); // ~ 7 MB of data
+    protected static final int FLUSH_SIZE_SMALL = 30; // ~ 3 MB (less than PART_SIZE, trigger single part upload - during commit)
+    protected static final int FLUSH_SIZE_BIG = 70; // ~ 7 MB (more than PART_SIZE, trigger two part uploads - during write and commit)
+    protected static final int FLUSH_SIZE_HUGE = 1400; // ~ 140 MB (more than 128 MB (default block size for parquet), trigger two part uploads for parquet)
+
+    protected static EmbeddedConnectCluster connect;
```
cjolivier01, line 59: These tests are great!

cjolivier01:
@lion-simba FYI: https://github.com/lion-simba/kafka-connect-storage-cloud/pull/1

lion-simba:
Hi @cjolivier01,

Thanks for your feedback!

I reviewed lion-simba#1 - please take a look.

As for `S3_RETRY_MAX_BACKOFF_TIME_MS` vs `max.poll.interval`, I don't think increasing `max.poll.interval` is a good idea, because it may slow down rebalances (for example, if a partition needs to be revoked from a slow task, coordinator has to wait up to max.poll.interval before it can forcibly remove that task from a group and reassign the partition). After this PR is accepted, instead of making `S3_RETRY_MAX_BACKOFF_TIME_MS` configurable, I would rather make it dependent on max.poll.interval setting (i.e. max time spent in retries handled by AWS SDK should be less than configured value of max.poll.interval). There are some corner cases though, so I suggest to think about it in a scope of a different PR.

lion-simba:
Hi @cjolivier01,

I merged your PR. CI check is the same - two failing tests are unrelated to this PR.

What will be our next step to get this PR approved and merged?

cjolivier01:
> Hi @cjolivier01,
> 
> I merged your PR. CI check is the same - two failing tests are unrelated to this PR.
> 
> What will be our next step to get this PR approved and merged?

Sounds good.  Can you change the base target to 10.0.x branch rather than master?  That will get it deployed to confluent hub.  I did a test at this and looks like you'll just need to adjust some of the version-related stuff, such as in the pom.  After that, I will ask IT to trigger the CI job and then we can merge it in.  Thanks so much for this fantastic contribution!

lion-simba:
@cjolivier01,

> Can you change the base target to 10.0.x branch rather than master?

Done.

The automatic check seems fine (same two unrelated tests are failing).

>  After that, I will ask IT to trigger the CI job and then we can merge it in.

Sounds good! Thanks!

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
```diff
@@ -0,0 +1,104 @@
...
+ * Utilities related to basic S3 error/exception analysis.
+ */
+public class S3ErrorUtils {
+
+  /**
+   * Return whether the given exception is a "retryable" exception.
+   * @param exception The exception to analyze
+   * @return true if the exception is retryable
+   */
+  private static boolean isRetryableException(Throwable exception) {
```
kpatelatwork, line 37: To be consistent with other methods in the class and RetriableException, we should rename this also to `isRetriableException` instead of `isRetryableException`

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -358,7 +357,7 @@ public Long getOffsetToCommitAndReset() {
   }
 
   public Long currentStartOffset() {
-    return currentStartOffset;
+    return minStartOffset();
```
kpatelatwork, line 360: this feels like doing two fixes in the same PR. should this fix and its tests go to a separate PR?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
```diff
@@ -0,0 +1,104 @@
...
+    if (exception == null) {
+      return false;
+    }
+    if (exception instanceof IOException) {
+      if (exception.equals(exception.getCause())) {
```
kpatelatwork, line 42: Instead of recursion should we use [ExceptionUtils.html#getRootCause](https://commons.apache.org/proper/commons-lang/javadocs/api-3.4/org/apache/commons/lang3/exception/ExceptionUtils.html#getRootCause(java.lang.Throwable))  as the current code may not handle case where its E1->E2->E3->E1

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
```diff
@@ -0,0 +1,104 @@
...
+      // Exception: An IOException embedded within an `AmazonClientException`
+      // should be passed via the `AmazonClientException` object
+      // as its parent (as the SDK does), in which case, shouldRetry()
+      // will often find it retryable.
+      return isRetryableException(exception.getCause());
+    }
+    if (exception instanceof AmazonClientException) {
+      // The AWS SDK maintains a check for what it considers to be
+      // retryable exceptions.
+      return PredefinedRetryPolicies.DEFAULT_RETRY_CONDITION.shouldRetry(
```
kpatelatwork, line 58: Nice

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -219,19 +219,26 @@ private MultipartUpload newMultipartUpload() throws IOException {
...
-        // and that retrying this request will not result in a successful response. This includes
-        // errors such as incorrect access keys, invalid parameter values, missing parameters, etc.
-        // Therefore, the connector should propagate this exception and fail.
-        throw new ConnectException("Unable to initiate MultipartUpload", e);
-      }
-      throw new IOException("Unable to initiate MultipartUpload.", e);
+      return supplier.get();
     } catch (AmazonClientException e) {
-      throw new IOException("Unable to initiate MultipartUpload.", e);
+      throw new IOException(e);
```
cjolivier01, line 241: should we be passing a message down here in order to not lose "Unable to initiate MultipartUpload." ?

cjolivier01:
I think there's several things that we can explore in the future wrt S3SinkTask::put(), including:
* Do we want to keep retrying each partition?  Do we have any reason to believe the others will succeed when, for instance, it's a throttling exception?  Continuing to hammer S3 would be counter-productive in this case.
* Upon  one (or more? all?) retried partitions, should we throw RetriableException back to kafka to do the exponential backoff?  This would be similar to workflow step backoff logic in workflow engines such as AWS SWF, Airflow, Cadence, etc.
* Other possibilities?

kpatelatwork:
@cjolivier01 and @lion-simba the fix has the right intent but some things that feels odd. It seems there were two attempts to make this fix in the past and we may be going in circles 

1. https://github.com/confluentinc/kafka-connect-storage-cloud/pull/284/files
2. https://github.com/confluentinc/kafka-connect-storage-cloud/pull/378/files

Some things that look odd are:-

1. We are trying to do `S3ErrorUtils.maybeRetriableConnectException(e)` in many places, this doesn't seem future proof.
2. In S3OutputStream we are wrapping exceptions only to unwrap them later from many places. Instead, should we centralize all the logic in S3OutputStream and throw RetriableException from there or centralize the logic in TopicPartitionWriter?

@cjolivier01 Let's also discuss this offline to see if we can do anything better here.

kpatelatwork:
> I think there's several things that we can explore in the future wrt S3SinkTask::put(), including:
> 
> * Do we want to keep retrying each partition?  Do we have any reason to believe the others will succeed when, for instance, it's a throttling exception?  Continuing to hammer S3 would be counter-productive in this case.
> * Upon  one (or more? all?) retried partitions, should we throw RetriableException back to kafka to do the exponential backoff?  This would be similar to workflow step backoff logic in workflow engines such as AWS SWF, Airflow, Cadence, etc.
> * We should probably log number of retries per put if > 1 into info or warn
> * Other possibilities?

do you mind logging a  ticket with labels "tech-debt" and "S3" so we don't lose these ideas as we should do some of these in a separate PR?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
lion-simba: Please see my response to @raphaelauv above: https://github.com/confluentinc/kafka-connect-storage-cloud/pull/461#issuecomment-989554722 He had the same concern as you.

Shortly, without other changes from this PR, there are no problems with this method, so no need to fix it separately.

This change is covered by integration tests added in this PR. Do you want me to also add a unit test for this?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
lion-simba: I removed the message intentionally because now this method is used in two places - during initiating multipart upload and during completion of multipart upload. So keeping the message would be misleading. I decided that having original exception passed as a cause to IOException will be enough to figure out what was the origin of the exception.

Do you think we still need to pass some message to IOException in addition to a "cause" exception?

lion-simba:
> Some things that look odd are:-
> 1. We are trying to do S3ErrorUtils.maybeRetriableConnectException(e) in many places, this doesn't seem future proof.
> 2. In S3OutputStream we are wrapping exceptions only to unwrap them later from many places. Instead, should we centralize all the logic in S3OutputStream and throw RetriableException from there or centralize the logic in TopicPartitionWriter?

These sound reasonable. My original intention was to make as minimum changes to the code (not tests) as possible to fix the issue. But if it's a blocker for a merge - let me know, I'll figure out something.

lion-simba:
> It seems there were two attempts to make this fix in the past and we may be going in circles
> 1. https://github.com/confluentinc/kafka-connect-storage-cloud/pull/284/files
> 2. https://github.com/confluentinc/kafka-connect-storage-cloud/pull/378/files

There is more, actually. For example, here is a PR which was closed by the author: https://github.com/confluentinc/kafka-connect-storage-cloud/pull/423/files

But I don't think these was circles, I think these were steps towards a proper solution:

https://github.com/confluentinc/kafka-connect-storage-cloud/pull/423/files was incomplete (not all places fixed)

https://github.com/confluentinc/kafka-connect-storage-cloud/pull/284/files added main infrastructure code for such kind of retries to work, but messed with exception types by counting unretriable exceptions as retriable:

```
    } catch (ConnectException e) {
      throw new RetriableException(e);
    }
```

https://github.com/confluentinc/kafka-connect-storage-cloud/pull/378/files fixed the problem of the previous PR and made a move into right direction, but it still was incomplete - the fix was only applied to commit() methods, but wasn't applied to write() methods of corresponding RecordWriterProvider's.

Now, this PR should hopefully fix the issue completely. A lot of tests in this PR serve this particular goal - make sure that all places are finally fixed (actually, it was a failing test, which pointed me to a problem with minStartOffset()).

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
lion-simba: Fixed.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
lion-simba: Fixed.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
```diff
@@ -0,0 +1,107 @@
...
+        // The AWS SDK maintains a check for what it considers to be
+        // retriable exceptions.
+        return PredefinedRetryPolicies.DEFAULT_RETRY_CONDITION.shouldRetry(
+                AmazonWebServiceRequest.NOOP,
+                (AmazonClientException) cause,
+                Integer.MAX_VALUE
+        );
+      }
+
+      if (!(cause instanceof IOException)) {
```
kpatelatwork, line 62: Every exception in the chain would not be IOException. Should we remove the if block as this is an exceptional situation and it's ok to iterate over all of them and outside the loop we are anyway returning false?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
kpatelatwork: Thanks @lion-simba for the explanation, it now makes sense to keep in this PR. 

If you can add a test it would be great :).

kpatelatwork:
> Now, this PR should hopefully fix the issue completely. A lot of tests in this PR serve this particular goal - make sure that all places are finally fixed (actually, it was a failing test, which pointed me to a problem with minStartOffset()).

Thanks, @lion-simba Agree with your assessment and reducing the amount of code changes, let me spend more time on the code and after holidays discuss offline with @cjolivier01.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
```diff
@@ -0,0 +1,107 @@
...
+
+    // IOException, in many places, is passed the AWS exception
+    // when it is thrown. We search here to check that exception
+    // for the IOException case.  Otherwise, the IOException
+    // is considered not retriable.
+    // Exception: An IOException embedded within an `AmazonClientException`
+    // should be passed via the `AmazonClientException` object
+    // as its parent (as the SDK does), in which case, shouldRetry()
+    // will often find it retriable.
+    for (Throwable cause : ExceptionUtils.getThrowableList(exception)) {
```
cjolivier01, line 51: @lion-simba I am not following why you replaced the recursion here with a loop.  The logic seems different now.  Can you explain?

cjolivier01: I don't think it is correct to dig more than one level deep into an AWS exception for an IOException.  The SDK does not do this: https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedRetryPolicies.java#L192

cjolivier01: nevermind, this is ok

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
lion-simba: I'm okay to remove it. However, the logic will be different than in recursive version, originally proposed by @cjolivier01. Please discuss with him offline and let me know your decision.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
lion-simba: I just rewrote the original recursive version in a loop form with better "cyclic reference" handling (as suggested above by @kpatelatwork). The logic should have remained exactly the same. Please let me know if it hasn't.

However, @kpatelatwork is suggesting to change original logic so that not only IOException is inspected for AmazonClientException cause (see [above](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/461#discussion_r775041556)). Please discuss with him and let me know your decision.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
lion-simba: I was about to add a unit test, but suddenly realized that tests defined in `io.confluent.connect.s3.TopicPartitionWriterTest` are actually kind of **integration tests** for TopicPartitionWriter: they communicate with mocked S3 instance, which is similar to how my tests in this PR work. So, if I add a new test for this particular code change to TopicPartitionWriterTest in a similar way, then this new test will be pretty much a duplicate to my tests added in this PR. And it makes a little sense to me to do so.

It looks like there is no other test types for TopicPartitionWriter exist. So, maybe it is okay to have only tests added in this PR? What do you think?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
kpatelatwork: We discussed and I think Chris has raised a PR against your branch. 

It's holidays here now @lion-simba so we will discuss it in the new year.

lion-simba: Oh, I didn't notice the PR. Reviewed now - @cjolivier01, please take a look.

Yeah, it's holidays here too. :) Wish you all happy New Year!

lion-simba:
Hi @cjolivier01 and @kpatelatwork,

I've merged PR prepared by Chris as well as updated 10.0.x branch.

Tests in "Jenkins-public-CI" are passing (except two tests which are failing due to CI misconfiguration). And I don't have permissions to check why "continuous-integration/jenkins/pr-merge" is failing.

What are next steps to get this PR merged?

lion-simba:
Hi @cjolivier01 and @kpatelatwork,

Any updates on this?

kpatelatwork:
Hey @lion-simba please bear with us a bit more we are also interested in merging this PR.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -366,7 +365,7 @@ public Long getOffsetToCommitAndReset() {
...
```
ncliang, line 368: I am not sure I understand the need for this change. From your description
> Otherwise, if IOException happens during partUpload() but before first call to commit(), Connector doesn't rewind kafka offset back in S3SinkTask.java:239, which leads to a data loss.

If that were to happen, wouldn't `currentStartOffset` not be advanced due to not calling commit? So the rewinding in `S3SinkTask.java:239` should rewind to previous `currentStartOffset` , right? Unless we have a path that would advance the offsets without calling commit, which seems like it would cause data loss as you say, regardless of whether the exception was retriable or not. That seems like it would indicate a bigger problem in the connector. For instance, if the connector were to be restarted in the middle of `partUpload()` , before `commit()` , would that cause data loss?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
```diff
@@ -76,7 +76,7 @@ public void write(SinkRecord record) {
             writer.setCodec(CodecFactory.fromString(conf.getAvroCodec()));
             writer.create(avroSchema, s3out);
           } catch (IOException e) {
-            throw new ConnectException(e);
+            throw S3ErrorUtils.maybeRetriableConnectException(e);
```
ncliang, line 79: nit: I think that the `maybeRetriableConnectException` being a factory method for creating an exception is slightly misleading. What do you think about making the method construct and throw the correct exception and just calling it something like `throwConnectException` ?

ncliang (commented):
Thanks, @lion-simba . I have just one question about the change and a minor nit. Otherwise, LGTM. The test changes are great!

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
lion-simba: Hi @ncliang,

> If that were to happen, wouldn't `currentStartOffset` not be advanced due to not calling commit? So the rewinding in `S3SinkTask.java:239` should rewind to previous `currentStartOffset` , right?

In original code, `currentStartOffset == null` before first call to `commit()`, so in `S3SinkTask.java:239` no rewinding is happening due to failed null check just above. Then, in `S3SinkTask.java:242` new `TopicPartitionWriter` is created, which start processing from where the previous `TopicPartitionWriter` left off (because no rewinding had taken place). So the records consumed by the previous `TopicPartitionWriter` get lost (because they were neither committed to S3 by old `TopicPartitionWriter` nor re-processed by a new `TopicPartitionWriter`).

> Unless we have a path that would advance the offsets without calling commit, which seems like it would cause data loss as you say, regardless of whether the exception was retriable or not.

No, there is no such path. The problem only appears if lines `S3SinkTask.java:242` and `S3SinkTask.java:244` are executed (replacing existing `TopicPartitionWriter` with a new one). And these lines are located inside of `catch (RetriableException e)` block, so they can execute only if we get an instance of `RetriableException`.

> For instance, if the connector were to be restarted in the middle of `partUpload()` , before `commit()` , would that cause data loss?

No, it wouldn't. On restart, all in-memory state got reset. Connector will start consuming from the latest offsets committed to Kafka. And these offsets will include all "uncommitted to S3" records, because offsets are only committed to Kafka after successful commit to S3 (`TopicPartitionWriter.java:610`).

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
lion-simba: Well, I did the change, and it looks fine by me, except one place:
In `JsonRecordWriterProvider.java:126` I had to add a dummy `return null` statement, because otherwise compiler complains about "missing return statement". Which is not a big deal I think. What do you think?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
kpatelatwork: Thanks, let me check if existing tests suffice.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
```diff
@@ -102,7 +102,7 @@ public void commit() {
           s3out.commit();
           writer.close();
         } catch (IOException e) {
-          throw new RetriableException(e);
+          throwConnectException(e);
         }
       }
```
kpatelatwork, line 108: @lion-simba and @cjolivier01 any reasons why the below `close()` method directly throws ConnectException instead of `throwConnectException(e)`, should this also be retriable?

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TestWithMockedS3.java
```diff
@@ -76,7 +77,9 @@
   @Override
   public void setUp() throws Exception {
     super.setUp();
-    port = url.substring(url.lastIndexOf(":") + 1);
+    if (StringUtils.isBlank(port)) {
```
kpatelatwork, line 80: Should we add a comment here telling if an explicit port is already set by child testcase do not overwrite it?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
cjolivier01: appears fixed

kpatelatwork: @cjolivier01 please check line 114 in this class as shown in screenshot I do not see it is fixed :).
image

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java
kpatelatwork: @lion-simba I talked with @cjolivier01 internally and he is ok with removing this. please remove this.

kpatelatwork:
@lion-simba  the changes look good and have excellent test coverage. Thanks a lot for the changes.

@cjolivier01 do you think you can take this PR over and resolve the  two minor nits I sent. Then its ready for final review and approval.

cjolivier01 (approved):
Thanks so much for this and sorry that it took so long!

cjolivier01:
ran all tests and verified no breakage.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
lion-simba: @kpatelatwork,
I didn't change it in first place, because there is no place in code, which can handle RetriableException thrown from here. But it shouldn't harm either.

lion-simba:
@kpatelatwork , @cjolivier01 , @ncliang ,
Thank you all for working on it with me!

PR #468, opened 2021-12-22
CCMSG-1582 set the next scheduled rotation duration even if there are…

kpatelatwork:
… no records in buffer and time has passed

## Problem
If there are no records in the buffer and no open files and 1000 new records are pushed to a topic at the same time then should these 1000 records be part of a single new file.
Currently, the task creates 2 new files (1 with 1st record and another with rest of the 999 records). This happens because we reset the `nextScheduledRotation` only if we have records to commit.

## Solution

We should reset the `nextScheduledRotation` even if there are no records to commit and time has passed

- [X] 
- [ ] no

##### If yes, where?
GCS and AzureBlob sink connector

##### Testing done:
- [X ] Unit tests
- [X] Integration tests
- [ ] System tests
- [X] Manual tests

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -331,6 +331,8 @@ public class TopicPartitionWriter {
         setNextScheduledRotation();
 
         commitFiles();
+      } else if (recordCount == 0 && shouldApplyScheduledRotation(now)) {
+        setNextScheduledRotation();
       }
 
       resume();
@@ -396,14 +398,18 @@ public class TopicPartitionWriter {
     );
 
     log.trace(
-        "Checking rotation on time with recordCount '{}' and encodedPartition '{}'",
+        "Checking rotation on time for topic-partition '{}' "
+                + "with recordCount '{}' and encodedPartition '{}'",
+        tp,
         recordCount,
         encodedPartition
     );
 
     log.trace(
-        "Should apply periodic time-based rotation (rotateIntervalMs: '{}', baseRecordTimestamp: "
+        "Should apply periodic time-based rotation for topic-partition '{}':"
+            +    " (rotateIntervalMs: '{}', baseRecordTimestamp: "
             + "'{}', timestamp: '{}', encodedPartition: '{}', currentEncodedPartition: '{}')? {}",
+        tp,
         rotateIntervalMs,
         baseRecordTimestamp,
         recordTimestamp,
@@ -412,16 +418,23 @@ public class TopicPartitionWriter {
         periodicRotation
     );
 
+    boolean scheduledRotation = shouldApplyScheduledRotation(now);
+    return periodicRotation || scheduledRotation;
+  }
+
+  private boolean shouldApplyScheduledRotation(long now) {
     boolean scheduledRotation = rotateScheduleIntervalMs > 0 && now >= nextScheduledRotation;
     log.trace(
-        "Should apply scheduled rotation: (rotateScheduleIntervalMs: '{}', nextScheduledRotation:"
-            + " '{}', now: '{}')? {}",
-        rotateScheduleIntervalMs,
-        nextScheduledRotation,
-        now,
-        scheduledRotation
+            "Should apply scheduled rotation for topic-partition '{}':"
+                    + " (rotateScheduleIntervalMs: '{}', nextScheduledRotation:"
+                    + " '{}', now: '{}')? {}",
+            tp,
+            rotateScheduleIntervalMs,
+            nextScheduledRotation,
+            now,
+            scheduledRotation
     );
-    return periodicRotation || scheduledRotation;
+    return scheduledRotation;
   }
 
   private void setNextScheduledRotation() {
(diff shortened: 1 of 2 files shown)


cjolivier01 (approved):
Thanks for fixing this!  Looks good, one minor nit.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -331,6 +331,8 @@ private void commitOnTimeIfNoData(long now) {
         setNextScheduledRotation();
 
         commitFiles();
+      } else if (recordCount == 0 && shouldApplyScheduledRotation(now)) {
```
cjolivier01, line 334: Do you need to call shouldApplyScheduledRotation(now) here?  Why not always call setNextScheduledRotation() for setNextScheduledRotation?  This might be just about to expire when the first one comes in, and then the same thing will happen -- split after too few records?

kpatelatwork: We need to check against now because some people might have the setting as 1 hour and they will complain I sent records at 11:59 but they are appearing after an hour.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -396,32 +398,44 @@ private boolean rotateOnTime(String encodedPartition, Long recordTimestamp, long
     );
 
     log.trace(
-        "Checking rotation on time with recordCount '{}' and encodedPartition '{}'",
```
kpatelatwork, line 399: added topic-partition to all these logs customer sent logs and without this it was not easy to figure out which log is for which partition as customer had many topics and many partitions on a connector.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
kpatelatwork: Also put with empty records list could be called 100s of time within a ScheduledRotation so we do not want to reset it on every call.

cjolivier01: I don;t have a strong opinion about it, but I am a little confused.  Why wouldn't you want setNextScheduledRotation() called every time there's no records received yet?  It's a relatively inexpensive call.  I was thinking that since these schedules are relative times (durations), they'd want the clock to start ticking at the first record written so that each file is written to for the same amount of time.

kpatelatwork:
@cjolivier01 and @kkonstantine  I had to tweak the solution a bit based on an edge case I found in manual testing. It's ready for another round of review.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -396,32 +404,44 @@ private boolean rotateOnTime(String encodedPartition, Long recordTimestamp, long
...
         baseRecordTimestamp,
         recordTimestamp,
         encodedPartition,
         currentEncodedPartition,
         periodicRotation
     );
+    if (periodicRotation) {
+      return true;
+    }
+    return shouldApplyScheduledRotation(now);
```
kkonstantine, line 429: Any reason not to write this as follows?
```suggestion
    return periodicRotation || shouldApplyScheduledRotation(now);
```

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -599,6 +599,167 @@ public void testWallclockUsesBatchTimePartitionBoundary() throws Exception {
...
+    // Bring the clock to present.
+    time.sleep(SYSTEM.milliseconds());
+
+    TopicPartitionWriter topicPartitionWriter = new TopicPartitionWriter(
+        TOPIC_PARTITION, storage, writerProvider, partitioner, connectorConfig, context, time,
+        null);
+
+    // sleep for 11 minutes after startup
+    time.sleep(TimeUnit.MINUTES.toMillis(11));
+    //do not reset the ScheduleRotation by calling topicPartitionWriter.write() but send records
```
kkonstantine, line 635: Is this comment still correct? I see write being called below

kkonstantine (commented):
Thanks @kpatelatwork . That's a nice fix. 
Just a couple of nits and also I'm wondering, does this port easily to 5.4.x and 5.5.x?

kkonstantine (approved):

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
kpatelatwork: Good suggestion I fixed it.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
kpatelatwork: Thanks, I agree this was unclear. I clarified the comments in both tests.

kpatelatwork:
@kkonstantine Thanks a lot for spending time on this. I will check if it's easily portable to 5.4.x and 5.5.x and get back to you.

Issue #389, opened 2021-01-27
Error while using SSE-C on AWS


Koisell:
Version: 5.5.3
While configuring SSE-C I faced the following error `com.amazonaws.services.s3.model.AmazonS3Exception: Server Side Encryption with Customer provided key is incompatible with the encryption method specified`.
After some investigation, it appears to reproduce the error described by mcsio in this [post (5th message)](https://forums.aws.amazon.com/thread.jspa?messageID=811033&t ).
Here are some DEBUG logs I got:
```
Jan 26 17:55:04 db1 connect-standalone.sh[27930]: [2021-01-26 17:55:04,563] DEBUG http-outgoing-12 >> "x-amz-server-side-encryption: AES256[\r][\n]" (org.apache.http.wire:73)
Jan 26 17:55:04 db1 connect-standalone.sh[27930]: [2021-01-26 17:55:04,563] DEBUG http-outgoing-12 >> "x-amz-server-side-encryption-customer-algorithm: AES256[\r][\n]" (org.apache.http.wire:73)
Jan 26 17:55:04 db1 connect-standalone.sh[27930]: [2021-01-26 17:55:04,563] DEBUG http-outgoing-12 >> "x-amz-server-side-encryption-customer-key: ****************=[\r][\n]" (org.apache.http.wire:73)
Jan 26 17:55:04 db1 connect-standalone.sh[27930]: [2021-01-26 17:55:04,563] DEBUG http-outgoing-12 >> "x-amz-server-side-encryption-customer-key-MD5: x/F0oeLbYckXF8ksG+dksA==[\r][\n]" (org.apache.http.wire:73)
```
Followed by
```
Jan 26 17:55:04 db1 connect-standalone.sh[27930]: [2021-01-26 17:55:04,592] DEBUG http-outgoing-12 << "<Error><Code>InvalidArgument</Code><Message>Server Side Encryption with Customer provided key is incompatible with the encryption method specified</Message><ArgumentName>x-amz-server-side-encryption</ArgumentName><ArgumentValue>AES256</ArgumentValue><RequestId>505A846706B4FE0E</RequestId><HostId>Es+YRbHxnKS6L1jZ2S0k3g3+O79lNq1bPbIgk0totJQlt3mlWaFZgRq1xhaIBpd27b5FNPdke38=</HostId></Error>[\r][\n]" (org.apache.http.wire:73)
```

[...]

PR #477, opened 2022-01-13
fix: removed x-amz-server-side-encryption to get SEE-C working

2phost:
## Problem
SSE-C is not working due to ```x-amz-server-side-encryption``` header. See  #389 .

## Solution
Simple fix, just remove that header. S3 just do not support having  ```x-amz-server-side-encryption``` and ```x-amz-server-side-encryption-customer-algorithm``` set at the same time

- [X] yes
- [ ] no

## Test Strategy
I tested it in v5.5.7 that is the version that I am using in my cluster, but it is the same for all versions.


2phost:
Note that I didn't test the KMS encryption because it is impossible to test it in my environment.
If for some reason it do not work, I can add this meta only for this case.

PR #480, opened 2022-01-13
Introduces the concept for source topic/partition in the TopicPartitionWritter

fmeyer:
## Problem

Whenever the destination topic is modified by any transformer, we loose the
ability to pause the original source topic, therefore it also checks against
isTransformedTopic when pausing/resuming topics during writes/flushes using the
record's original topic/partition.

## Solution

This should fix any transform NPE that happens when a transform that modifies the topic name is present

Introduces the concept for source topic in the TopicPartitionWritter builded from the information contained in the InternalSinkRecord. 

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -295,12 +295,9 @@ public class S3SinkTask extends SinkTask {
 
     for (TopicPartition tp : topicPartitionWriters.keySet()) {
       TopicPartitionWriter topicPartitionwritter = topicPartitionWriters.get(tp);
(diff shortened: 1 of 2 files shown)


snehashisp:
The inability to use transformations that change the underlying topic name is amongst the limitations on the [S3 connectors](https://docs.confluent.io/kafka-connect-s3-sink/current/overview.html#limitations). The limitations are there because, in addition to proper functioning of the pause/resume functions, the worker relies on the offsets sent back by the connector to ensure delivery guarantees. If the topic name are changed the offsets returned may not be the right offset (with the correct topic name) to be committed.

btiernay:
@snehashisp Thanks for the explanation. What's the likelihood of this PR being accepted?

PR #489, opened 2022-02-07
MINOR: Suppress warning which causes IntelliJ to error after Maven import (via -Werror)

cjolivier01:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


kpatelatwork (approved):
@cjolivier01 could we please target 10.0.x branch as that's where most changes happen.

LGTM and Thanks for this change.

PR #492, opened 2022-02-08
CCMSG-1481 Retriable exception is possible to be thrown from close()

cjolivier01:
## Problem

Did some debugging and verified that it is, in fact, possible for a retriable exception to be thrown during close().  Currently it appears that this won't be retried if it's during a task close(), but that's another matter (that maybe should be addressed separately).

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


kpatelatwork (approved):
LGTM

cjolivier01:
Having some sort of network error, will retry later (Gateway timeout)

PR #493, opened 2022-02-11
Generalize the S3 retry mechanism via a simple wrapper class.

cjolivier01:
## Problem

Consolidte the catch/rethrow logic for AWS retriable exceptions by wrapping the anonymous format classes as they are created.  Wrapper is intentionally as close as possible to the call/throw sites rather than up another level (or more) in KeyValueHeaderRecordWriterProvider.

No behavioral changes should occur.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


cjolivier01:
moving to my fork

PR #495, opened 2022-02-25
Don't log full proxy config, don't log configuration @ info level

cjolivier01:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


kpatelatwork (approved):
LGTM

PR #497, opened 2022-03-07
Fix regions list being one item of a list

cjolivier01:
## Problem

validValues() returns singleton list of a list of 27 regions, making the single-selectable item a list of 27 items. It is impossible to set a region in CP due to this.

image

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


BlueEldur (approved):
LGTM

PR #509, opened 2022-04-21
CCMSG-1833 Remove the logging of sink record in TRACE message 

kpatelatwork:
## Problem
SinkRecord was being logged in TRACE message exposing confidential data

## Solution
only log the topic+partition+offset for the record instead of contents

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


BlueEldur (approved):
LGTM

PR #516, opened 2022-05-26
CCMSG-1577 : label change to compression level for s3 sink connector

akash-borkar:
## Problem
Jira ticket : https://confluentinc.atlassian.net/browse/CCMSG-1577
The label for Compression Level config is incorrect in the S3 Sink connector for Confluent Platform

## Solution

In the S3 sink connector Confluent Platform the label for the config 's3.compression.level' is incorrectly set as 'Compression Type' . Look at the screen shot attached which shows how the config label looks in Confluent Platform.

We must rename that label to 'Compression Level'

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


pbadani (approved):
base changed to 10.0.x. LGTM

Issue #520, opened 2022-06-03
Write a lot of small files in s3 when there are two versions schema data in topic


harveyyue:
Currently, we capture 100 shard tables to a topic in debezium. DBA add four columns to these shard tables one by one, this means there are two version schema in this topic. Unfortunately, dba met the lock table issue, they paused the rest of tables schema change action, we had to write a lot of small files in s3 folder and got the huge lag in kafka consumer processing.

The s3 small files:
2022-06-03 03:26:17   46.4 KiB db_name.db_user_data.tb_all_shards+0+8138464479.snappy.parquet
2022-06-03 03:26:17   31.9 KiB db_name.db_user_data.tb_all_shards+0+8138464550.snappy.parquet
2022-06-03 03:26:17   36.7 KiB db_name.db_user_data.tb_all_shards+0+8138464551.snappy.parquet
2022-06-03 03:26:17   33.1 KiB db_name.db_user_data.tb_all_shards+0+8138464573.snappy.parquet
2022-06-03 03:26:18   54.7 KiB db_name.db_user_data.tb_all_shards+0+8138464575.snappy.parquet
2022-06-03 03:26:18   30.1 KiB db_name.db_user_data.tb_all_shards+0+8138464670.snappy.parquet
2022-06-03 03:26:18   38.3 KiB db_name.db_user_data.tb_all_shards+0+8138464672.snappy.parquet
2022-06-03 03:26:18   39.9 KiB db_name.db_user_data.tb_all_shards+0+8138464693.snappy.parquet
2022-06-03 03:26:19   30.1 KiB db_name.db_user_data.tb_all_shards+0+8138464733.snappy.parquet
2022-06-03 03:26:19   34.8 KiB db_name.db_user_data.tb_all_shards+0+8138464734.snappy.parquet
2022-06-03 03:26:19   40.5 KiB db_name.db_user_data.tb_all_shards+0+8138464736.snappy.parquet
2022-06-03 03:26:19   34.8 KiB db_name.db_user_data.tb_all_shards+0+8138464770.snappy.parquet
2022-06-03 03:26:20   51.4 KiB db_name.db_user_data.tb_all_shards+0+8138464772.snappy.parquet
2022-06-03 03:26:20   35.3 KiB db_name.db_user_data.tb_all_shards+0+8138464852.snappy.parquet
2022-06-03 03:26:20   47.7 KiB db_name.db_user_data.tb_all_shards+0+8138464856.snappy.parquet
[...]

PR #521, opened 2022-06-03
Write a lot of small files in s3 when there are two versions schema data in topic

harveyyue:
## Problem
https://github.com/confluentinc/kafka-connect-storage-cloud/issues/520
## Solution
Modify the encoded partition with schema version to support the different schema.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


viswa-optimusprime:
@harveyyue @blaisdellk @ijuma @blaisdellk , We need to this feature. Can you commit files. We have use case where dont want to rotate file when schema or schema version change in same topic .

viswa-optimusprime:
When we can expect the change?

PR #537, opened 2022-07-29
To fix the CVE CCDB-4985 in kafka-connect-databricks-delta-lake

devinay:
## Problem
The CVE in kafka-connect-databricks-delta-lake  : https://confluentinc.atlassian.net/browse/CCDB-4985
Needs the version of aws-java-sdk bumped up.
This library derives from kafka-connect-s3, which comes from this repository.
The problem is this `pom.xml` file does not contain a version entry for aws-java-sdk and therefore cannot be pinned in the dependent pom in kafka-connect-databricks-delta-lake.

## Solution
Add the version entry so that it can be overridden and pint merge to the latest.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/kafka-connect-s3/pom.xml
+++ b/kafka-connect-s3/pom.xml
@@ -36,7 +36,7 @@
     </description>
 
     <properties>
-        <aws.version>1.11.725</aws.version>
+        <aws.version>1.12.270</aws.version>
         <s3mock.version>0.1.5</s3mock.version>
         <apache.httpclient.version>4.5.9</apache.httpclient.version>
         <kafka.connect.maven.plugin.version>0.11.1</kafka.connect.maven.plugin.version>
--- a/kafka-connect-s3/pom.xml
+++ b/kafka-connect-s3/pom.xml
@@ -65,6 +65,7 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk-s3</artifactId>
+            <version>${aws.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -175,7 +175,7 @@
                   <groupId>org.apache.maven.plugins</groupId>
                   <artifactId>maven-surefire-plugin</artifactId>
                   <configuration>
-                      <argLine>@{argLine} -Djava.awt.headless=true -Xmx1024m -XX:MaxPermSize=512m</argLine>
+                      <argLine>@{argLine} -Djava.awt.headless=true -XX:MaxPermSize=512m</argLine>
                       <reuseForks>false</reuseForks>
                       <forkCount>1</forkCount>
                   </configuration>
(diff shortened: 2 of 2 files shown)


anupamaggarwal:
Might be slighty more safer to update the version coming from the bom by updating the version of the bom here - https://github.com/confluentinc/kafka-connect-storage-cloud/blob/master/kafka-connect-s3/pom.xml? 
Although it might entail more regressions 
I think this is the version you want (https://repo1.maven.org/maven2/com/amazonaws/aws-java-sdk-bom/1.12.270/aws-java-sdk-bom-1.12.270.pom)

devinay:
> Might be slighty more safer to update the version coming from the bom by updating the version of the bom here - https://github.com/confluentinc/kafka-connect-storage-cloud/blob/master/kafka-connect-s3/pom.xml? Although it might entail more regressions I think this is the version you want (https://repo1.maven.org/maven2/com/amazonaws/aws-java-sdk-bom/1.12.270/aws-java-sdk-bom-1.12.270.pom)

Did you mean updating this same file's master branch ? That would still leave regressions open in the older releases so I'm trying to set it to `1.12.270` in a slightly older branch.

The issue is that the `aws-java-sdk-s3` dependency [does not have](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/1ab539fe6662519ba91942d34a86aca14aa7fd16/kafka-connect-s3/pom.xml#L70) a `aws.version` that can be overridden in a dependant pom downstream.

Is there anything I am missing ?

anupamaggarwal:
I meant it would probably be safer to simply update the ${aws.version} (like you have already) in this repo  and let the bom import all specific versions of aws artifacts (this way the bom can ensure compatibility/resolve conflicting versions). You won't need to set an explicit version here ,this would also entail extra maintenance overhead. Once you deploy the updated version  of kafka-connect-s3 in repo you can then update the the delta-lake repo to pull in this updated verion of kafka-connect-s3.  Alternatively, you should also be able to pin a direct dependency in delta lake repo (although is potentially unsafe since multiple aws libs (1.11.x and 1.12.x) might not interoperate well.....is this not working?

devinay:
> Alternatively, you should also be able to pin a direct dependency in delta lake repo (although is potentially unsafe since multiple aws libs (1.11.x and 1.12.x) might not interoperate well.....is this not working?

No, pinning it downstream in data lake by setting `<aws.version>1.12.270</aws.version>` does not work.

devinay:
> I meant it would probably be safer to simply update the ${aws.version} (like you have already) in this repo and let the bom import all specific versions of aws artifacts (this way the bom can ensure compatibility/resolve conflicting versions). 

Yes that would work for now, but in general for future CVE's, say the case of databricks delta lake for example, there is no dependency on aws-java-sdk-bom, so its not possible to specifically override `aws-java-sdk-s3`. This is why the change to explicitly set versions.

PR #551, opened 2022-08-17
Upgrade s3 sdk to 1.12.268

xjin-Confluent:
## Problem
This is to upgrade S3 sdk version

- [x] yes
- [ ] no

##### If yes, where?
CP

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/kafka-connect-s3/pom.xml
+++ b/kafka-connect-s3/pom.xml
@@ -35,7 +35,7 @@
     </description>
 
     <properties>
-        <aws.version>1.11.1015</aws.version>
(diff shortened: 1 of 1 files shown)


snehashisp (approved):
LGTM. Please update the title of the PR by adding the version of aws we are updating to. Also we dont need to include 5.4 and 5.5 in the title. Being picky as we auto generate changelog updated from the commit name (which defaults to the title) 😄.

PR #554, opened 2022-09-03
CCDB-4880: Adding a S3 Sink Connector config to decide whether to ignore object tagging errors or not.

ManasjyotiSharma:
## Problem
[CCDB-4880](https://confluentinc.atlassian.net/browse/CCDB-4880): Databricks Delta Lake Sink Connector sporadically fails with a NPE when trying to read "endOffset" tag from S3 objects. The root cause is that the Delta Lake Sink Connector under the hood leverages S3 Sink Connector code to write/create S3 objects and the S3 Sink Connector code in case of any tagging errors simply ignores them. However from Databricks Sink Connector POV ignoring such tagging errors means the S3 object is left in an inconsistent state because it expects that if object is there in S3 then tag must also be there for that S3 object. See discussions in this [slack thread](https://confluent.slack.com/archives/C011CCECDB7/p1660623278820309).

## Solution
Adding a S3 Sink Connector low level configuartion so that the client/caller can decide whether it is okay for it to ignore tagging errors or not. This PR is part-1 of a 2 parts PR. The part-2 PR will have the Databricks connector side of changes leveraging the newly added S3 Sink config & handle any tagging errors gracefully at Databricks Sink connector end. However until this current S3 Sink PR is merged and a new version is created, I cannot raise part-2 PR as in Databricks Sink Connector pom I'd have to use the new version number of S3 Sink Connector.

**PS:** Note that even though in the slack thread mentioned above, I mentioned about adding a retry logic, however that is already there within the S3 Sink code, see [S3Storage#newFullJitterRetryPolicy](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/10.0.x/kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3Storage.java#L160) method. Since the [TopicPartitionWriter#tagFile](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/10.0.x/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java#L658) method leverages an instance of S3Storage, so the S3 tagging error should have been retried 3 times with 200 ms exponential backoff (doubling each time). Since the original issue occurred sometime back, I am no longer able to grab the logs from Kibana/OpenSearch to verify whether that S3Storage retry logic got kicked in or not. However that aspect is not specific to the current Delta Lake issue and I'll leave it to the S3 Sink Connector owner to explore later in terms of general retry logic of S3Storage class implementation.

- [x] yes
- [ ] no

##### If yes, where?
Databricks Delta Lake Sink Connector

## Test Strategy
Manually tested via docker playground the overall tagging failure scenario from Databricks connector end by hardcoding a throw clause in the S3 Sink Connector's TopicPartitionWriter#tagFile method.

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -80,6 +80,9 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
   public static final String S3_OBJECT_TAGGING_CONFIG = "s3.object.tagging";
   public static final boolean S3_OBJECT_TAGGING_DEFAULT = false;
 
+  public static final String S3_OBJECT_IGNORE_TAGGING_ERROR_CONFIG = "ignore.tagging.errors";
+  public static final boolean S3_OBJECT_IGNORE_TAGGING_ERROR_DEFAULT = true;
+
   public static final String SSEA_CONFIG = "s3.ssea.name";
   public static final String SSEA_DEFAULT = "";
 
@@ -269,6 +272,18 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
           ++orderInGroup,
           Width.LONG,
           "S3 Object Tagging"
+      );
+
+      configDef.define(
+              S3_OBJECT_IGNORE_TAGGING_ERROR_CONFIG,
+              Type.BOOLEAN,
+              S3_OBJECT_IGNORE_TAGGING_ERROR_DEFAULT,
+              Importance.LOW,
+              "Whether to ignore any S3 objects tagging error or not.",
+              group,
+              ++orderInGroup,
+              Width.LONG,
+              "Ignore S3 Object Tagging Error"
       );
 
       configDef.define(
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -80,8 +80,10 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
   public static final String S3_OBJECT_TAGGING_CONFIG = "s3.object.tagging";
   public static final boolean S3_OBJECT_TAGGING_DEFAULT = false;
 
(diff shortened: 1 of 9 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -271,6 +274,18 @@ public static ConfigDef newConfigDef() {
           "S3 Object Tagging"
       );
 
+      configDef.define(
+              S3_OBJECT_IGNORE_TAGGING_ERROR_CONFIG,
+              Type.BOOLEAN,
```
pbadani, line 279: This could be modelled as behavior.on.tagging.error (along the lines of behavior.on.null.values) with 2 options ignore, and fail.

pbadani:
Since this is a new feature, a minor version bump is more appropriate as opposed to a patch update.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -659,10 +662,18 @@ private void tagFile(String encodedPartition, String s3ObjectPath) {
       log.info("Tagged S3 object {} with starting offset {}, ending offset {}, record count {}",
           s3ObjectPath, startOffset, endOffset, recordCount);
     } catch (SdkClientException e) {
-      log.warn("Unable to tag S3 object {}. Ignoring.", s3ObjectPath, e);
+      if (ignoreTaggingErrors) {
+        log.warn("Unable to tag S3 object {}. Ignoring.", s3ObjectPath, e);
```
pbadani, line 666: Another way to avoid consuming errors would be to adjust the retry policy, passing the behaviour on tagging config and deciding there whether to retry.

sudeshwasnik: +1 , we could use - https://github.com/confluentinc/kafka-connect-storage-cloud/blob/21da6c93b7108a26060d55895b7261f48246f990/kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/S3ErrorUtils.java#L15-L86

pbadani: this was just a mention.

PR #578, opened 2022-11-03
bumping hadoop and jettison version

pkuntalteam1:
## Problem
[CCMSG-2287](https://confluentinc.atlassian.net/browse/CCMSG-2287) - Upgrade jettison at kafka-connect-storage-cloud
[CCMSG-2293](https://confluentinc.atlassian.net/browse/CCMSG-2293) - Upgrade org.apache.hadoop:hadoop-common at kafka-connect-storage-cloud

## Solution
1- modifying hadoop-common version to  3.2.3.
2- And adding jettison dependency with version 1.5.1 explicitly.

<--- Snippet from dependency listing--->

[INFO] +- commons-beanutils:commons-beanutils:jar:1.9.4:compile
[INFO] +- org.codehaus.jettison:jettison:jar:1.5.1:compile
[INFO] +- org.apache.hadoop:hadoop-mapreduce-client-core:jar:3.2.3:test
[INFO] |  +- org.apache.hadoop:hadoop-yarn-client:jar:3.2.3:test
[INFO] |  |  \- org.apache.hadoop:hadoop-yarn-api:jar:3.2.3:test

+- org.apache.hadoop:hadoop-common:jar:3.2.3:compile
[INFO] |  +- org.apache.hadoop:hadoop-annotations:jar:3.2.3:compile

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,11 +131,16 @@
             <artifactId>jackson-core</artifactId>
             <version>${jackson.version}</version>
         </dependency>
+
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>${hadoop.version}</version>
+            <version>3.2.3</version>
             <exclusions>
+                <exclusion>
+                    <groupId>com.sun.jersey</groupId>
+                    <artifactId>jersey-json</artifactId>
+                </exclusion>
                 <exclusion>
                     <groupId>org.apache.avro</groupId>
                     <artifactId>avro</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -131,11 +131,16 @@
...
             <version>${jackson.version}</version>
         </dependency>
+
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>${hadoop.version}</version>
+            <version>3.2.3</version>
             <exclusions>
+                <exclusion>
```
snehashisp, line 140: any reason why we are excluding here and pinning the version below, if we want to change the version hadoop is using we should put it under dependency management

pom.xml
```diff
@@ -197,6 +202,22 @@
             <artifactId>commons-beanutils</artifactId>
             <version>1.9.4</version>
         </dependency>
+        <dependency>
+            <groupId>com.sun.jersey</groupId>
+            <artifactId>jersey-json</artifactId>
+            <version>1.19</version>
+            <exclusions>
```
snehashisp, line 209: same as above comment, we can also just put this under dependency management

snehashisp:
A couple of notes
* Don't think the tests are getting executed. 
* Can we also target a later branch and check if tests are are successful on those branches.

pom.xml
```diff
@@ -131,6 +133,7 @@
             <artifactId>jackson-core</artifactId>
             <version>${jackson.version}</version>
         </dependency>
+
```
pbadani, line 136: nit: can remove this change

pbadani (approved):

janjwerner-confluent:
@pkuntalteam1 and @snehashisp 
thank you for updating this, unfortunately 3.2.3 is already obsolete and hitting the scanners with a new CVE, could you please bump it one up to 3.2.4

PR #594, opened 2022-11-11
fixing release issue

pkuntalteam1:
## Problem
while release, it is picking SNAPshot version.

## Solution
Updating it with proper version

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


sudeshwasnik (approved):
CI build is not running. 
Approving since it was built locally.

PR #603, opened 2022-12-08
CCMSG-2368: Cache short-lived credentials when using AWSAssumeRoleCredentialProvider

sidd1809:
## Problem
[CCMSG-2368](https://confluentinc.atlassian.net/browse/CCMSG-2368): Some customers reported seeing 427 (throttles) from AWS STS service because we are not caching STS assume role call's credentials. 

## Solution
- Start caching  credentials fetched vis STSAssumeRoleCredential provider
- AWS sdk's STSAssumeRoleCredentialProvider implementation already takes care of refreshing the fetched temp creds via a background [thread](https://github.com/aws/aws-sdk-java/blob/6059d1033b8be1de544f679732e18220d22a848a/aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSAssumeRoleSessionCredentialsProvider.java#L34-L48) before expiry

- [ ] yes
- [X] no

## Test Strategy
Tested manually for CP

##### Testing done:
- [X] Unit tests
- [X] Integration tests
- [ ] System tests
- [X] Manual tests

## Release Plan
Release a patch version

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
@@ -108,7 +108,6 @@ public class AwsAssumeRoleCredentialsProvider implements AWSCredentialsProvider,
 
   @Override
   public void refresh() {
-    // STSAssumeRoleSessionCredentialsProvider refreshes credentials every 900s, but refresh()
     // performs a force refresh of credentials
     if (stsCredentialProvider != null) {
       stsCredentialProvider.refresh();
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
@@ -67,6 +67,10 @@ public class AwsAssumeRoleCredentialsProvider implements AWSCredentialsProvider,
 
   private BasicAWSCredentials basicCredentials;
 
+  // STSAssumeRoleSessionCredentialsProvider takes care of refreshing short-lived
+  // credentials 60 seconds before it's expiry
+  private STSAssumeRoleSessionCredentialsProvider stsCredentialProvider;
+
   @Override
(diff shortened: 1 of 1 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
```diff
@@ -77,32 +81,38 @@ public void configure(Map<String, ?> configs) {
...
-            .build()
-            .getCredentials();
-    }
+    return stsCredentialProvider.getCredentials();
   }
 
   @Override
   public void refresh() {
-    // Nothing to do really, since we acquire a new session every getCredentials() call.
+    // STSAssumeRoleSessionCredentialsProvider refreshes credentials every 900s, but refresh()
```
sidd1809, line 111:
```suggestion
    // STSAssumeRoleSessionCredentialsProvider refreshes credentials every 900s, but refresh()
```

kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
```diff
@@ -77,32 +81,37 @@ public void configure(Map<String, ?> configs) {
...
-            .build()
-            .getCredentials();
-    }
+    return stsCredentialProvider.getCredentials();
   }
 
   @Override
   public void refresh() {
-    // Nothing to do really, since we acquire a new session every getCredentials() call.
+    // performs a force refresh of credentials
```
pkuntalteam1, line 111: Do we also need to refresh basicCredential also?

pkuntalteam1 (approved):
LGTM.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/auth/AwsAssumeRoleCredentialsProvider.java
sidd1809: No we don't need to refresh that, since basicCredentials if set (i.e provided in connector config) are not ephemeral in this case. 

If basic credentials was null, the sts credential provider would use default credentials provider chain that should automatically refresh the creds provided from underlying host...

PR #651, opened 2023-05-30
CC-19464 | Add logs to track the count for each of the rotation criteria. This helps debug at runtime the number of rotations that occurred for each criteria.

pbadani:
## Problem
Currently when file rotations happen, we do not know the reason behind them. In this PR we print the rotation counts for each output file so that we know th e

- [ ] yes
- [x] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


aniketshrimal (approved):
LGTM @pbadani! One minor comment - we can add some unit tests for newly added class - FileRotationTracker.

BDeus:
Sorry to come late on this. It could be nice to expose this new metrics as JMX to monitor the number of files write by connectors.

PR #660, opened 2023-08-09
ESC-581: pin dependencies to resolve CVEs for Citi escalation

janjwerner-confluent:
## Problem
Following up on https://confluentinc.atlassian.net/browse/ESC-581
this is plugging all the currently known CVEs for Citi connector ingestion. 
##### Does this solution apply anywhere else?
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

PR #680, opened 2023-09-11
Fix for regression from CVE fix in PR #660

ashoke-cube:
## Problem
Excluding `hadoop-shaded-guava` for CVE fix in commit https://github.com/confluentinc/kafka-connect-storage-cloud/commit/8a80a19363f5c8d89d7b52367352a25b1ff1de8b
resulted in a regression with `NoClassDefFoundError: org/apache/hadoop/thirdparty/com/google/common/collect/Interners`
while using the `io.confluent.connect.s3.format.parquet.ParquetFormat` for `format.class` field in config.

## Solution
[kafka-connect-storage-common](https://github.com/confluentinc/kafka-connect-storage-common/tree/master) has an updated hadoop-shaded-guava without the CVE.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/kafka-connect-s3/src/assembly/package.xml
+++ b/kafka-connect-s3/src/assembly/package.xml
@@ -66,14 +66,5 @@
                 <exclude>io.confluent:kafka-connect-storage-partitioner</exclude>
             </excludes>
         </dependencySet>
-        <dependencySet>
-        <outputDirectory>share/java/kafka-connect-s3</outputDirectory>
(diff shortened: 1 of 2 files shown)


pbadani (approved):
LGTM after testing with parquet input format in docker playground.

PR #745, opened 2024-04-24
CC-24623 Handle SchemaParseException and InvalidSchemaException according to errors.tolerance

subhashiyer9:
## Problem
In case of any malformed record failing due to SchemaParseException or InvalidSchemaException, connector fails with an error instead of adhering to errors.tolerance parameter

## Solution
Catch the related exceptions and write the error to dlq topic based on the configured dlq and errors.tolerance

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


aniketshrimal (approved):
LGTM @subhashiyer9! Just checking - do we need to update our public documentation that these exceptions will be routed to dlq? If yes, please do followup with docs ticket.

subhashiyer9:
> LGTM @subhashiyer9! Just checking - do we need to update our public documentation that these exceptions will be routed to dlq? If yes, please do followup with docs ticket.

We don't explicitly document the various errors that will go to dlq, so I think it should be fine as it is since this falls under the category of malformed records

Issue #770, opened 2024-08-08
kafka-connect-s3 SSE without customer key broken


85danf:
Versions:
- 10.5.6
- 10.4.8
- 10.3.10
- 10.2.14
- 10.1.15
- 10.0.26

Hey, we've been happily using kafka-connect-s3 for a long time and recently after a version upgrade one of our usecases started to break:
It's a bucket not owned by my team which has `AES256` SSE configured with an aws-provided key (not customer provided).
up until the version upgrade it was working perfectly and since the upgrade we started getting permission errors like 

```java
Caused by: com.amazonaws.services.s3.model.AmazonS3Exception: User: arn:aws:sts::***:assumed-role/***/aws-sdk-java-*** is not authorized 
to perform: kms:GenerateDataKey on this resource because the resource does not exist in this Region, no resource-based policies allow access, 
or a resource-based policy explicitly denies access (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: ***; 
S3 Extended Request ID: ***; Proxy: null), S3 Extended Request ID: ***
```

I've pinpointed the issue to this [commit](https://github.com/confluentinc/kafka-connect-storage-cloud/commit/10b020aa3e8f7cc12cf5bfb2661a2a9fe4470cfe) / [PR](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/684).

Looks like its not taking this usecase under consideration and always expects a customer key to be available, which is something that is taken into account when initializing the customer key in `S3OutputStream::79` : 
```java
this.sseCustomerKey = (SSEAlgorithm.AES256.toString().equalsIgnoreCase(ssea)
        && StringUtils.isNotBlank(sseCustomerKeyConfig))
      ? new SSECustomerKey(sseCustomerKeyConfig) : null;
```

I think the `x-amz-server-side-encryption` header still has to be included when `ssea == AES256` and `sseCustomerKey == null`

PR #772, opened 2024-08-08
Add x-amz-server-side-encryption header for AES256 SSE without customer key

85danf:
## Problem
See #770 

## Solution
The code in this PR adds the `x-amz-server-side-encryption` header in cases where `SSE` is configured to be `AES256` but not customer key was provided.

- [X] yes
- [ ] no

##### If yes, where?
The following versions (and onwards) are affected, I'm not sure about this repo's patch strategy so I opened it on the latest branch.
- 10.5.6
- 10.4.8
- 10.3.10
- 10.2.14
- 10.1.15
- 10.0.26

## Test Strategy
Built locally and tested on our environment.

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [X] Manual tests


subhashiyer9:
@85danf Thanks for the fix.
Can you rebase to the oldest affected branch 10.0.x and update the PR?

PR #793, opened 2024-10-22
CC-29553 Fix connector to work with SSE-S3

subhashiyer9:
## Problem
The connector does not add `x-amz-server-side-encryption` header when SSE is set to AES256 and no customer key is specified. This results in connector failing in cases where the IAM policy expects `x-amz-server-side-encryption` to be set.

## Solution
Add the SSE related metadata to init request when customer key is not null

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/kafka-connect-s3/src/test/java/io/confluent/connect/s3/storage/S3OutputStreamTest.java
+++ b/kafka-connect-s3/src/test/java/io/confluent/connect/s3/storage/S3OutputStreamTest.java
@@ -1,33 +1,47 @@
 package io.confluent.connect.s3.storage;
 
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import com.amazonaws.AmazonClientException;
 import com.amazonaws.AmazonServiceException;
 import com.amazonaws.AmazonServiceException.ErrorType;
 import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.model.InitiateMultipartUploadRequest;
+import com.amazonaws.services.s3.model.InitiateMultipartUploadResult;
+import com.amazonaws.services.s3.model.SSEAlgorithm;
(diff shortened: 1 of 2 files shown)


kafka-connect-s3/src/test/java/io/confluent/connect/s3/storage/S3OutputStreamTest.java
```diff
@@ -53,4 +67,77 @@ public void testPropagateOtherRetriableS3Exceptions() {
     when(s3Mock.initiateMultipartUpload(any())).thenThrow(new AmazonClientException(S3_EXCEPTION_MESSAGE));
     assertThrows(IOException.class, () -> stream.commit());
   }
+
+  @Test
+  public void testNewMultipartUploadAESSSE() throws IOException {
```
sp-gupta, line 72: Is AES with customer key, a valid case? If yes, we should add a test for it as well.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/storage/S3OutputStreamTest.java
```diff
@@ -53,4 +67,77 @@ public void testPropagateOtherRetriableS3Exceptions() {
...
+    assertNull(captor.getValue().getSSEAwsKeyManagementParams());
+  }
+
+  @Test
+  public void testNewMultipartUploadKMSSSE() throws IOException {
+
+    Map<String, String> props = createProps();
+    props.put(S3SinkConnectorConfig.SSEA_CONFIG, SSEAlgorithm.KMS.toString());
+    props.put(S3SinkConnectorConfig.SSE_KMS_KEY_ID_CONFIG, "key1");
+    stream = new S3OutputStream(S3_TEST_KEY_NAME, new S3SinkConnectorConfig(props), s3Mock);
```
sp-gupta, line 95: Is KMS without key a valid case? If yes, we should add a test for it as well.

sp-gupta (approved):

kafka-connect-s3/src/test/java/io/confluent/connect/s3/storage/S3OutputStreamTest.java
subhashiyer9: KMS algorithm will need an associated KMS Id to work, other wise it will result in error.
But, possibly a good validation to include

PR #801, opened 2024-11-28
CC-29764 Add timeout to sink put cycle

subhashiyer9:
## Problem
This PR adds a time out to the put loop to add a limit on maximum amount of time a task can spend in the poll cycle
If the task spends significant amount of time in poll loop, it will likely become a zombie.
Hence, adding a timeout can prevent a zombie task from processing the files

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

Manually tested by adding random sleep in the task and ensuring that the task reaches deadline in such cases

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -554,7 +554,7 @@ public class TopicPartitionWriter {
   private void resume() {
     log.trace("Resuming writer for topic-partition '{}'", tp);
     context.resume(tp);
-    isPaused = true;
+    isPaused = false;
   }
 
   private RecordWriter newWriter(SinkRecord record, String encodedPartition)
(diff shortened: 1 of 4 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -502,11 +548,13 @@ private boolean rotateOnSize() {
   private void pause() {
     log.trace("Pausing writer for topic-partition '{}'", tp);
     context.pause(tp);
+    isPaused = true;
   }
 
   private void resume() {
     log.trace("Resuming writer for topic-partition '{}'", tp);
     context.resume(tp);
+    isPaused = true;
```
sp-gupta, line 557: This should be `false` right?

subhashiyer9: Thanks for pointing out. I have fixed it

sp-gupta (approved):

PR #844, opened 2025-05-27
Update File rotation based on a limit when FieldPartitioner is selected.

ahmedashfaq027:
## Problem

* When FieldPartitioner is selected, we do not want to overload the resources in Cloud and cause a failure.

JIRA: https://confluentinc.atlassian.net/browse/CC-33220

## Solution

* We're limiting Part size to 5MB, FieldPartitioner flush size to 500 files, and 1 Topic Partition per task.
* As part of this approach, we're limiting the flush size to 500 via template values in this commit.

- [ ] yes
- [x] no

## Test Strategy

* Tested by deploying the changes on devel canary-6. The limit is being applied correctly based on the configured value.

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -250,6 +250,9 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
       + "to prefix or suffix in the s3 path after the topic name."
       + " None will not append the schema name in the s3 path.";
 
+  public static final String FIELD_PARTITIONER_FLUSH_SIZE_CONFIG = "field.partitioner.flush.size";
+  public static final int FIELD_PARTITIONER_FLUSH_SIZE_DEFAULT = -1;
+
   private static final GenericRecommender SCHEMA_PARTITION_AFFIX_TYPE_RECOMMENDER =
       new GenericRecommender();
 
@@ -913,6 +916,12 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
           "Elastic buffer initial capacity"
       );
 
+      configDef.defineInternal(
+          FIELD_PARTITIONER_FLUSH_SIZE_CONFIG,
+          Type.INT,
+          FIELD_PARTITIONER_FLUSH_SIZE_DEFAULT,
+          Importance.LOW
+      );
     }
     return configDef;
   }
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
@@ -23,6 +23,7 @@ import io.confluent.connect.s3.util.FileRotationTracker;
 import io.confluent.connect.s3.util.RetryUtil;
 import io.confluent.connect.s3.util.TombstoneTimestampExtractor;
 import io.confluent.connect.storage.errors.PartitionException;
+import io.confluent.connect.storage.partitioner.FieldPartitioner;
 import io.confluent.connect.storage.schema.SchemaCompatibilityResult;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.data.Schema;
@@ -88,6 +89,7 @@ public class TopicPartitionWriter {
   private final boolean ignoreTaggingErrors;
   private int recordCount;
   private final int flushSize;
+  private final int fieldPartitionerFlushSize;
   private final long rotateIntervalMs;
   private final long rotateScheduleIntervalMs;
   private long nextScheduledRotation;
@@ -169,6 +171,8 @@ public class TopicPartitionWriter {
             S3SinkConnectorConfig.S3_OBJECT_BEHAVIOR_ON_TAGGING_ERROR_CONFIG)
             .equalsIgnoreCase(S3SinkConnectorConfig.IgnoreOrFailBehavior.IGNORE.toString());
     flushSize = connectorConfig.getInt(S3SinkConnectorConfig.FLUSH_SIZE_CONFIG);
+    fieldPartitionerFlushSize = connectorConfig.getInt(
+        S3SinkConnectorConfig.FIELD_PARTITIONER_FLUSH_SIZE_CONFIG);
     topicsDir = connectorConfig.getString(StorageCommonConfig.TOPICS_DIR_CONFIG);
     rotateIntervalMs = connectorConfig.getLong(S3SinkConnectorConfig.ROTATE_INTERVAL_MS_CONFIG);
     if (rotateIntervalMs > 0 && timestampExtractor == null) {
@@ -428,6 +432,17 @@ public class TopicPartitionWriter {
       return false;
     }
 
+    if (rotateOnFieldPartitioner()) {
+      fileRotationTracker.incrementRotationByFieldPartitionerCount(encodedPartition);
+      log.info(
+          "Starting commit and rotation for topic partition {} with start offset {}",
+          tp,
+          startOffsets
+      );
+      nextState();
+      return true;
+    }
+
     if (rotateOnSize()) {
       fileRotationTracker.incrementRotationByFlushSizeCount(encodedPartition);
       log.info(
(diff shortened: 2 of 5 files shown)


kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -658,6 +659,80 @@ public void testWriteRecordTimeBasedPartitionRecordTimestampHoursOutOfOrder() th
...
+    // Define the partitioner
+    Partitioner<?> partitioner = new FieldPartitioner<>();
+    partitioner.configure(parsedConfig);
+
+    TopicPartitionWriter topicPartitionWriter = new TopicPartitionWriter(
+        TOPIC_PARTITION, storage, writerProvider, partitioner, connectorConfig, context, null);
+
+    String key = "key";
+    Schema schema = createSchema();
+    List<Struct> records = createRecordBatches(schema, 1, 5);
```
subhashiyer9, line 715: Can we add at `flush.size + 1` number of records to ensure, the file is created at the end of the test.
 We can then verify it as `verify(expectedFiles, 1000, schema, records)` later [here](https://github.com/confluentinc/kafka-connect-storage-cloud/pull/844/files#diff-5eb1b1881ea8f1e5e169e39bc69f877f4414a73f4f27cccd8b4eff5b658c17dbR733) instead of checking for no expected files with 0 records

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
```diff
@@ -658,6 +659,80 @@ public void testWriteRecordTimeBasedPartitionRecordTimestampHoursOutOfOrder() th
     verify(expectedFiles, 2, schema, records);
   }
 
+  @Test
+  public void testWriteRecordFieldBasedPartitionAndRotateOnFlushSize() throws Exception {
+    localProps.put(FIELD_PARTITIONER_FLUSH_SIZE_CONFIG, "1");
```
subhashiyer9, line 664: Can you also add a test FIELD_PARTITIONER_FLUSH_SIZE_CONFIG with a value more than 1?
And also generate records which can different values for the field being partitioned on instead of a single value `String.valueOf(16)`

subhashiyer9: Another test can be added to ensure that the FIELD_PARTITIONER_FLUSH_SIZE_CONFIG is honored even there are multiple fields configured?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -442,6 +457,18 @@ private boolean checkRotationOrAppend(
     return false;
   }
 
+  private boolean rotateOnFieldPartitioner() {
```
subhashiyer9, line 460: If I understand correctly, this will always result in the last file being written to contain a single record. Can we avoid that?.

subhashiyer9: I am thinking of a condition like `!commitFiles.contains(encodedPartition) && commitFiles.size() >= fieldPartitionerFlushSize)`. This way we will try to rotate when we are about to create a new file .
Accordingly, we will also need to move the rotateOnFieldPartitoner before invoking the `writeRecord`

subhashiyer9 (changes requested):

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -250,6 +250,9 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
       + "to prefix or suffix in the s3 path after the topic name."
       + " None will not append the schema name in the s3 path.";
 
+  public static final String FIELD_PARTITIONER_FLUSH_SIZE_CONFIG = "field.partitioner.flush.size";
```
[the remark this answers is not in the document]
ahmedashfaq027: sure, but we wanted to do this only on field partitioner for now, isn't it?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/FileRotationTracker.java
```diff
@@ -97,6 +103,13 @@ public void incrementRotationByFlushSizeCount(String outputPartition) {
     metrics.get(outputPartition).incrementRotationByFlushSizeCount();
   }
 
+  public void incrementRotationByFieldPartitionerCount(String outputPartition) {
```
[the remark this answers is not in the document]
ahmedashfaq027: This increment would be called when the rotation returns true (Commit files > 500). So we're not tracking it per encoded partitions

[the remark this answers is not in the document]
subhashiyer9: Yes, but the increment method is invoked on the current encodedPartition, which seems incorrect to me

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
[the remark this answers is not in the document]
ahmedashfaq027: Yes, Makes sense. We can do that!

kafka-connect-s3/src/test/java/io/confluent/connect/s3/TopicPartitionWriterTest.java
ahmedashfaq027: Updated the tests

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -442,6 +457,18 @@ private boolean checkRotationOrAppend(
...
+    if (!(partitioner instanceof FieldPartitioner) || partitionerMaxOpenFiles == -1) {
```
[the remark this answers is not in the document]
ahmedashfaq027: Then we should be updating the method name?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
ahmedashfaq027: Okay, updated the condition back, and moved it before writing the record.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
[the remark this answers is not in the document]
ahmedashfaq027: This condition is necessary, because we're only enforcing the condition of max files, only if the partitioner class is FieldPartitioner

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
```diff
@@ -442,6 +457,19 @@ private boolean checkRotationOrAppend(
...
+      return false;
+    }
+
+    boolean rotate = !commitFiles.containsKey(currentEncodedPartition)
+        && commitFiles.size() > partitionerMaxOpenFiles;
```
subhashiyer9, line 466: This should be `commitFiles.size() == partitionerMaxOpenFiles`?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/TopicPartitionWriter.java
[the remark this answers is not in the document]
ahmedashfaq027: Removed this, and renamed the method.

This value is now controlled from templates

sp-gupta (commented):
We should change the base branch to a feature branch

kafka-connect-s3/src/main/java/io/confluent/connect/s3/util/FileRotationTracker.java
```diff
@@ -97,6 +103,13 @@ public void incrementRotationByFlushSizeCount(String outputPartition) {
     metrics.get(outputPartition).incrementRotationByFlushSizeCount();
   }
 
+  public void incrementRotationByPartitionerMaxOpenFilesCount(String outputPartition) {
+    if (!metrics.containsKey(outputPartition)) {
+      metrics.put(outputPartition, new RotationMetrics());
+    }
```
sp-gupta, line 109: nit: We can use SonarQube suggestion here

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -913,6 +916,12 @@ public static ConfigDef newConfigDef() {
           "Elastic buffer initial capacity"
       );
 
+      configDef.defineInternal(
```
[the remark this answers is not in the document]
ahmedashfaq027: We do not want user/UI to set this config, and Hence it's internal

subhashiyer9 (approved):

PR #948, opened 2026-07-24
feat(INIT-5113): add BACKUP_FULL_RECORD mode to S3 sink connector

sribaabu:
## Problem
  
  The S3 sink connector has no way to back up full Kafka records with schema metadata preservation. Users need long-term archival with replay capability, cross-environment data seeding, and schema registry migration support.
  
  ## Solution
  
  Add BACKUP_FULL_RECORD mode to S3 sink connector. Every record is wrapped in a self-describing envelope containing key, value, headers, topic, partition, offset, timestamp, and schema metadata. Schemas are backed up as pristine files
  alongside the data.
  
  - BackupS3SinkTask: extends S3SinkTask, wraps records via EnvelopeTransformer before super.put()
  - S3StorageWriter: writes schema files to S3 with S3ErrorUtils retry classification
  - S3SinkConnectorValidator: delegates to shared BackupModeValidator for config validation
  - S3SinkConnector.taskClass(): routes to BackupS3SinkTask for BACKUP_FULL_RECORD mode
  
  Also bumps parent to storage-common-parent:12.0.0-SNAPSHOT (common:8.4.0-1073) and fixes test infrastructure for Kafka 8.4.x: JUnit 5 @Tag migration, KRaft-based EmbeddedConnectCluster deps, Jetty 12 module rename, Scala 2.13 s3mock, Jakarta
  EE 10 EL, PowerMock --add-opens, test compile release=17.
  
  - [x] yes
  - [ ] no
  
  ##### If yes, where?
  GCS and Azure sink connectors follow the same pattern — separate PRs. Object-store-source (restore side, PR pending). Storage-common (shared logic, PR pending).
  
  ## Test Strategy
  
  ##### Testing done:
  - [x] Unit tests
  - [ ] Integration tests
  - [ ] System tests
  - [x] Manual tests
  
  11 backup round-trip tests covering Avro, Protobuf, JSON Schema, schema references, evolution, tombstones, enum preservation, and GUID-based schemas. 342/344 total tests pass (2 pre-existing AWS env failures). E2E validated via Docker on Connect 8.3.0 runtime with all schema types.
  
  ## Release Plan
  
  Merging to master via INIT-5113-phase2. Depends on storage-common PR merging first. Part of Backup & Restore feature (INIT-5113).

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/BackupS3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/BackupS3SinkTask.java
@@ -36,7 +36,7 @@ import java.util.Map;
  * before delegating to the parent {@link S3SinkTask}.
  *
  * <p>Records are transformed into envelope-shaped SinkRecords in
- * {@code put()} — before they reach {@code TopicPartitionWriter}.
+ * {@code put()}, before they reach {@code TopicPartitionWriter}.
  * This means all schema rotation decisions (key changes, value changes,
  * tombstone transitions) are based on the envelope schema, which
  * naturally handles every case without special-case logic.
@@ -58,9 +58,6 @@ import java.util.Map;
  *   <li>Byte-identical write if both prior levels miss (writes still
  *       succeed; last-write-wins produces the same bytes).</li>
  * </ol>
- *
- * <p>Follows the same pattern as the source side where
- * {@code RestoreS3SourceTask} extends the generic source task.
  */
 public class BackupS3SinkTask extends S3SinkTask {
 
@@ -94,6 +91,8 @@ public class BackupS3SinkTask extends S3SinkTask {
     BackupModeValidator.logSinkStartupSummary(
         originals, connectorConfig.formatClass().getSimpleName(),
         keyType, valueType);
+    log.info("Started backup-mode S3 sink task (keyType={}, valueType={}, format={})",
+        keyType, valueType, connectorConfig.formatClass().getSimpleName());
   }
 
   @Override
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/BackupS3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/BackupS3SinkTask.java
@@ -42,7 +42,22 @@ import java.util.Map;
  * naturally handles every case without special-case logic.
  *
  * <p>Schema files (.entry.json + .avsc/.proto/.json) are backed up
- * as a side effect during wrapping (idempotent, with 3-level dedup).
+ * as a side effect during wrapping. Writes are <b>idempotent</b>:
+ * the S3 path is keyed by schema ID and the content (raw schema bytes,
+ * entry.json fields) is a pure function of that key, so two tasks
+ * writing the same schema always produce identical bytes and S3
+ * last-write-wins is a no-op.
+ *
+ * <p>To avoid redundant work, {@link ObjectStoreSchemaBackupStore}
+ * applies <b>3-level deduplication</b>:
+ * <ol>
+ *   <li>In-memory {@code ConcurrentHashMap}-backed set of already-backed-up
+ *       schema keys (blocks re-attempts within the same task JVM).</li>
+ *   <li>S3 {@code HEAD} on the entry.json path (blocks cross-task
+ *       double-writes when another task has already completed).</li>
+ *   <li>Byte-identical write if both prior levels miss (writes still
+ *       succeed; last-write-wins produces the same bytes).</li>
+ * </ol>
  *
  * <p>Follows the same pattern as the source side where
  * {@code RestoreS3SourceTask} extends the generic source task.
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/BackupS3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/BackupS3SinkTask.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2025 Confluent Inc.
+ *
+ * Licensed under the Confluent Community License (the "License"); you may not use
+ * this file except in compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.confluent.io/confluent-community-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package io.confluent.connect.s3;
+
+import io.confluent.connect.s3.backup.S3StorageWriter;
+import io.confluent.connect.storage.backup.BackupEnvelope;
+import io.confluent.connect.storage.backup.BackupModeValidator;
+import io.confluent.connect.storage.backup.ConverterTypeDetector;
+import io.confluent.connect.storage.backup.ObjectStoreSchemaBackupStore;
+import io.confluent.connect.storage.backup.SchemaBackupStore;
+import io.confluent.connect.storage.common.StorageCommonConfig;
+import io.confluent.connect.storage.format.backup.EnvelopeTransformer;
+import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Backup-specific sink task that pre-wraps records in an envelope
+ * before delegating to the parent {@link S3SinkTask}.
+ *
+ * <p>Records are transformed into envelope-shaped SinkRecords in
+ * {@code put()} — before they reach {@code TopicPartitionWriter}.
+ * This means all schema rotation decisions (key changes, value changes,
+ * tombstone transitions) are based on the envelope schema, which
+ * naturally handles every case without special-case logic.
+ *
+ * <p>Schema files (.entry.json + .avsc/.proto/.json) are backed up
+ * as a side effect during wrapping (idempotent, with 3-level dedup).
+ *
+ * <p>Follows the same pattern as the source side where
+ * {@code RestoreS3SourceTask} extends the generic source task.
+ */
+public class BackupS3SinkTask extends S3SinkTask {
+
+  private static final Logger log = LoggerFactory.getLogger(BackupS3SinkTask.class);
+
+  private EnvelopeTransformer envelopeTransformer;
+
+  @Override
+  public void start(Map<String, String> props) {
+    super.start(props);
+
+    String topicsDir = connectorConfig.getString(
+        StorageCommonConfig.TOPICS_DIR_CONFIG);
+    String dirDelim = connectorConfig.getString(
+        StorageCommonConfig.DIRECTORY_DELIM_CONFIG);
+
+    SchemaBackupStore backupStore = new ObjectStoreSchemaBackupStore(
+        new S3StorageWriter(storage), topicsDir, dirDelim);
+
+    Map<String, String> originals = connectorConfig.originalsStrings();
+    String keyType = ConverterTypeDetector.detectSchemaType(
+        originals.get(BackupEnvelope.KEY_CONVERTER_CONFIG),
+        originals, BackupEnvelope.KEY_CONVERTER_CONFIG);
+    String valueType = ConverterTypeDetector.detectSchemaType(
+        originals.get(BackupEnvelope.VALUE_CONVERTER_CONFIG),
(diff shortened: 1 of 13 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -217,7 +219,7 @@ private Partitioner<?> newPartitioner(S3SinkConnectorConfig config)
     if (config.getSchemaPartitionAffixType() != S3SinkConnectorConfig.AffixType.NONE) {
       partitioner = new SchemaPartitioner<>(partitioner);
     }
-    if (config.isTombstoneWriteEnabled()) {
+    if (config.isTombstoneWriteEnabled() && !config.isBackupMode()) {
```
subhashiyer9, line 222: What happens when `config.isTombstoneWriteEnabled()` AND `config.isBackupMode()`?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -336,8 +338,8 @@ private boolean maybeSkipOnNullValue(SinkRecord record) {
         return false;
       } else {
         // Fail
-        throw new ConnectException("Null valued records are not writeable with current "
-            + S3SinkConnectorConfig.BEHAVIOR_ON_NULL_VALUES_CONFIG + " 'settings.");
+        throw new DataException("Null valued records are not writeable with current "
```
subhashiyer9, line 341: What is the reasoning behind this change?
IMO this is a change in behaviour.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
```diff
@@ -117,7 +116,7 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
-@Category(IntegrationTest.class)
+@org.junit.jupiter.api.Tag("integration")
```
subhashiyer9: What is the reason for this change?

kafka-connect-s3/pom.xml
```diff
@@ -142,15 +142,15 @@
...
-            <groupId>javax.el</groupId>
-            <artifactId>javax.el-api</artifactId>
-            <version>3.0.0</version>
+            <groupId>jakarta.el</groupId>
+            <artifactId>jakarta.el-api</artifactId>
+            <version>6.0.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.glassfish</groupId>
```
subhashiyer9, line 151: Curious to understand why did you need these test dependency changes? Is it because of the parent version bump?

kafka-connect-s3/pom.xml
```diff
@@ -428,7 +477,7 @@
                     </dependencies>
                     <configuration>
                       <skip>${skipIntegrationTests}</skip>
-                      <groups>org.apache.kafka.test.IntegrationTest</groups>
+                      <groups>integration</groups>
```
subhashiyer9: What is this change?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/backup/S3StorageWriter.java
```diff
@@ -0,0 +1,73 @@
...
+      LoggerFactory.getLogger(S3StorageWriter.class);
+
+  private final S3Storage storage;
+
+  public S3StorageWriter(S3Storage storage) {
+    this.storage = storage;
+  }
+
+  @Override
+  public void write(String path, String content) {
```
subhashiyer9, line 45: How does it safeguard against concurrent writes to same file from different tasks?
I remember the plan was to check etags and conditional writes to safeguard against this?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonFormat.java
```diff
@@ -37,7 +37,9 @@ public JsonFormat(S3Storage storage) {
     this.storage = storage;
     this.converter = new JsonConverter();
     Map<String, Object> converterConfig = new HashMap<>();
-    converterConfig.put("schemas.enable", "false");
+    boolean schemasEnable = storage.conf().isJsonSchemaEmbedded();
```
subhashiyer9, line 40: Why is this new config `isJsonSchemaEmbedded` needed for this feature?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/BackupS3SinkTask.java
```diff
@@ -0,0 +1,95 @@
...
+ * before delegating to the parent {@link S3SinkTask}.
+ *
+ * <p>Records are transformed into envelope-shaped SinkRecords in
+ * {@code put()} — before they reach {@code TopicPartitionWriter}.
+ * This means all schema rotation decisions (key changes, value changes,
+ * tombstone transitions) are based on the envelope schema, which
+ * naturally handles every case without special-case logic.
+ *
+ * <p>Schema files (.entry.json + .avsc/.proto/.json) are backed up
+ * as a side effect during wrapping (idempotent, with 3-level dedup).
```
subhashiyer9, line 45: nit: It will be helpful to explain `idempotent, with 3-level dedup` better for future reference.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorValidator.java
```diff
@@ -169,7 +187,7 @@ private void recordError(String message, String key) {
     Objects.requireNonNull(key);
     if (!key.equals("")) {
       ConfigValue value = valuesByKey.get(key);
-      if (!message.equals("")) {
+      if (value != null && !message.equals("")) {
```
subhashiyer9, line 190: QQ: Did you encounter any case where the value was null?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/backup/S3StorageWriter.java
```diff
@@ -0,0 +1,73 @@
...
+        s3out.close();
+      }
+      log.debug("Schema file written: {}", path);
+    } catch (Exception e) {
+      S3ErrorUtils.throwConnectException(e);
+    }
+  }
+
+  @Override
+  public boolean exists(String path) {
```
subhashiyer9, line 65: Is this simple file exists check sufficient or requires inspection of contents?

kafka-connect-s3/pom.xml
sribaabu: This was part of the CP version bump, Reverted.

kafka-connect-s3/pom.xml
```diff
@@ -408,12 +431,38 @@
...
+                    </execution>
+                  </executions>
+            </plugin>
             <plugin>
                   <groupId>org.apache.maven.plugins</groupId>
                   <artifactId>maven-surefire-plugin</artifactId>
                   <configuration>
                       <!-- Skip the Integration tests as part of surefire run -->
-                      <excludedGroups>org.apache.kafka.test.IntegrationTest</excludedGroups>
+                      <excludedGroups>integration</excludedGroups>
```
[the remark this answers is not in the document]
sribaabu: This was part of the CP version bump, Reverted.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/integration/BaseConnectorIT.java
sribaabu: This was part of the CP version bump, Reverted.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
sribaabu: This was a defensive code earlier where the TombstoneWrite earlier created separate /tombstone folder. But with the current approach the value is never null with the Envelope and null value is wrapped inside the envelope. Removing it.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
sribaabu: Agreed, reverted the changes.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/backup/S3StorageWriter.java
sribaabu: Old plan: single shared manifest.json per topic listing all schemas. Backup writes would need CAS/etag to safely merge concurrent updates; restore would GET and parse the whole manifest to look up one entry.
  
  Changed to: one {schemaId}.entry.json per schema (metadata) + one {schemaId}.{ext} per schema (raw schema bytes). Envelope already carries the schema ID, so restore computes the path directly and does a single S3 GET (O(1), no scan, no manifest parse). Writes are idempotent by content, so no CAS needed and parallel tasks never contend on the same file.

  Why: simpler correctness (no distributed compare-and-swap), no cross-task coordination, and cheaper reads on restore.

Ref - https://confluentinc.atlassian.net/wiki/spaces/~7120208d0cb1e7ae434cebae47b946ca37cc37/pages/5443223781/Backup+Restore+Schema+Backup+Restore+Design

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/json/JsonFormat.java
sribaabu: Backup wraps every record in an envelope Struct (key, value, headers, schema IDs, metadata). To restore, we need to reconstruct that Struct from the on-disk file, which requires the schema to be recoverable alongside the data.
  
Avro files embed the schema in the file header, Parquet files embed it in the footer. Restore reads the schema directly from the file itself, so no extra flag is needed.

JSON is not self-describing by default. JsonConverter writes payload-only JSON, so restore has no way to type the fields back. Turning on format.json.schema.enable=true switches it to {"schema":..., "payload":...}, bringing JSON to parity with Avro/Parquet for round-trip.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/BackupS3SinkTask.java
sribaabu: Updated.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorValidator.java
sribaabu: This was an obsolete change added to force explicitly setting the converter config, Reverted.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/backup/S3StorageWriter.java
sribaabu: exists() is enough here. Content is fully deterministic - path is keyed by schema ID and both the raw schema bytes and the entry.json fields are pure functions of that key, so two writers can never produce different content for the same path.  Nothing for a content check to catch. Also S3 multipart writes are atomic on commit, so the file only appears after the full write succeeds. Crashed mid-write leaves nothing visible and we just retry. A content check would mean a GET instead of a HEAD, more cost for no correctness.

kafka-connect-s3/pom.xml
sribaabu: It was due to the parent version bump, reverted.

subhashiyer9:
Overall LGTM.
Can you also add an integration test for this?

pom.xml
```diff
@@ -73,6 +73,8 @@
         <dnsjava.version>3.6.1</dnsjava.version>
         <commons.lang3.version>3.18.0</commons.lang3.version>
         <netty.version>4.2.16.Final</netty.version>
+        <!-- Override to consume locally-built storage-common with BACKUP_FULL_RECORD envelope classes -->
+        <kafka.connect.storage.common.version>12.1.0-SNAPSHOT</kafka.connect.storage.common.version>
```
subhashiyer9, line 77: Just a note to update this once the tag is available.

sribaabu: Sure.

kafka-connect-s3/src/test/java/io/confluent/connect/s3/BackupS3SinkTaskTest.java
```diff
@@ -0,0 +1,150 @@
...
+    putBackupModeProps();
+    setUp();
+    replayAll();
+    BackupS3SinkTask task = new BackupS3SinkTask();
+    task.initialize(context);
+    task.start(properties);
+    verifyAll();
+
+    // empty collection exercises wrap + super.put chain without SinkRecord setup
+    task.put(Collections.<SinkRecord>emptyList());
```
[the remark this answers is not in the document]
sribaabu: Covered in Unit test.

PR #950, opened 2026-08-07
Close only revoked partition writers in S3 sink task (KIP-848 partial revocation)

asehgal-confluent:
## Summary
- Under Kafka's new consumer group protocol (KIP-848), a task can be asked to revoke only a *subset* of its assigned partitions, keeping the rest. `S3SinkTask.close(Collection<TopicPartition>)` ignored its `partitions` argument and tore down **every** tracked writer, dropping buffered records for partitions the task still owned and was never going to re-open.
- Also lazily creates a writer if a record arrives for a partition that is already in the task's live assignment but whose writer hasn't been opened yet (a startup/rebalance timing race), instead of failing the task.
- Confirmed live on a real Confluent Cloud canary running this exact code path: `No writer found for topic partition ... does not match any assigned partitions` at `S3SinkTask.getTopicPartitionWriterOrThrow`, reproduced repeatedly on a plain multi-task connector.

## Test plan
- [x] Added `testPartialRevocationPreservesRetainedPartitionState` and `testPutRecordForPartitionInAssignmentButNotYetOpenedCreatesWriter` regression tests
- [x] `mvn compile test-compile` passes cleanly for `kafka-connect-s3`
- [ ] Could not execute the test suite locally — this repo's PowerMock 2.0.9-based tests are incompatible with JDK 17+ module restrictions, and no JDK 8/11 was available in this environment. Recommend running the existing + new tests in CI before merge.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -371,14 +371,18 @@ public class S3SinkTask extends SinkTask {
 
   @Override
   public void close(Collection<TopicPartition> partitions) {
-    for (TopicPartition tp : topicPartitionWriters.keySet()) {
-      try {
-        topicPartitionWriters.get(tp).close();
-      } catch (ConnectException e) {
-        log.error("Error closing writer for {}. Error: {}", tp, e.getMessage());
+    // Only close writers for the partitions actually being revoked; revocation can be partial
+    // and the retained partitions keep receiving records without a fresh open().
+    for (TopicPartition tp : partitions) {
+      TopicPartitionWriter writer = topicPartitionWriters.remove(tp);
+      if (writer != null) {
+        try {
+          writer.close();
+        } catch (ConnectException e) {
+          log.error("Error closing writer for {}. Error: {}", tp, e.getMessage());
+        }
       }
     }
-    topicPartitionWriters.clear();
   }
 
   @Override
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
@@ -276,17 +276,23 @@ public class S3SinkTask extends SinkTask {
   }
 
   /**
-   * Retrieves the TopicPartitionWriter for the given TopicPartition.
-   * If no writer is found, it indicates a mismatch between the record's topic partition
-   * and the assigned partitions.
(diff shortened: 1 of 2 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
```diff
@@ -276,17 +276,23 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
...
    *
    * @param tp the TopicPartition to get the writer for
    * @return the TopicPartitionWriter
-   * @throws ConnectException if no writer is found
+   * @throws ConnectException if the partition is not currently assigned to this task
    */
   private TopicPartitionWriter getTopicPartitionWriterOrThrow(TopicPartition tp)
       throws ConnectException {
     TopicPartitionWriter writer = topicPartitionWriters.get(tp);
+    if (writer == null && context.assignment().contains(tp)) {
```
subhashiyer9: Wouldn't this have already been created during rebalance on `open` method?

hisingla12: @subhashiyer9 Thanks for pointing it out. 

During rebalance - `onPartitionsAssigned()` does call the [task.open()](https://github.com/confluentinc/ce-kafka/blob/08812b7037da7510ad458c8cd32cf1c7fc1ccea1/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L1025) and it does create the writer on assignment before records arrive, so this branch isn't hit on the normal path. 
I think it was added (Jul 25) as the first fix for the `No writer found` crash, before the actual root cause - `close()` wiping all writers on a partial revocation - was found and fixed (Aug 7).  

@asehgal-confluent can you please confirm this

asehgal-confluent: Thats true - The close() fix alone is what matters here. 
open() change is cheap defensive fallback so thought of keeping it. But happy to strip off to keep the PR minimally scoped

hisingla12: Reverted the open() change to keep the PR scoped.

subhashiyer9 (approved):

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkTask.java
subhashiyer9: Can this be backported to 10.6.x onwards?

hisingla12:
Build is failing on trivy scan. All tests are passing.

PR #955, opened 2026-08-24
fix: keep sensitive data out of S3 sink connector logs

sumit-confluent:
## What

Sensitive-log audit fixes for the S3 sink connector (audited `kafka-connect-storage-cloud` `v12.1.10` @ `8bcd9479`, via the `kafka-connect-s3-private` wrapper). This PR ships the log / message-level fixes; the one config-typing change (F1) is reverted (see below).
**Report**: https://drive.google.com/file/d/12pacD5k6yweJJXOnCpP7inQjkFktA2PT/view?usp=drive_link

## Findings shipped (log / message-level)

| Finding | file:line | Fix | Test |
|---|---|---|---|
| F2 — encoded path in logs | `TopicPartitionWriter.java` (rotation INFO/WARN + `findNextAvailableFile`/`tagFile`, incl. the two tag-failure throws) | Log the topic-partition and offsets/counts instead of the `encodedPartition` / S3 object path | `TopicPartitionWriterTest#testFieldPartitionerDoesNotLogRecordFieldValues`, `testFailS3ObjectTagging{SdkClientException,RuntimeException}` |
| F3 — S3 object key in logs | `S3OutputStream.java` | Log only the `bucket`, not the full object key; key stays at DEBUG | `S3OutputStreamTest#testCreateLogDoesNotContainObjectKey` |
| F4 — S3 object path in logs | `S3Storage.java` | Log only the `bucket`, not the object path | `S3StorageTest#testCreateParquetStreamDoesNotLogObjectKey` |
| F5 — filename in logs | `AvroRecordWriterProvider.java` | Move the adjusted-filename line from INFO to DEBUG | covered by `testFieldPartitionerDoesNotLogRecordFieldValues` |
| F6 — filename in logs | `ParquetRecordWriterProvider.java` | Move the adjusted-filename line from INFO to DEBUG | `DataWriterParquetTest#testParquetRecordWriterDoesNotLogFilenameAtInfo` |
| F7 — proxy URL userinfo | `S3ProxyConfig.java` (`ConfigException` on malformed URL) | Redact `user:password@` userinfo from `s3.proxy.url` before it is echoed (`s3.proxy.url` stays `Type.STRING`) | `S3ProxyTest#testProxyUrlUserInfoRedactedInConfigException` |

## Deferred / reverted (config-typing)

- **F1** — `aws.access.key.id` — **reverted; stays `Type.STRING`.** Retyping to `PASSWORD` would mask it as `[hidden]` in the config dump, Connect UI and `/validate` response — an externally visible config-contract change. Per the decision to ship no config-contract changes this release, and the confirmation that `aws.access.key.id` is not classified sensitive, it stays plain `Type.STRING` (read via `getString()`). Deferred to a future release.

## Behavior

- No config-contract or breaking changes: `aws.access.key.id` and `s3.proxy.url` remain plain `Type.STRING`; no config is retyped and no validation is added.
- Only log lines / exception message content change (F2–F7).

## Testing

- `kafka-connect-s3` compiles; checkstyle clean. Affected tests pass (`S3ProxyTest`, `S3OutputStreamTest`, and the F2 field-partitioner tests). Three pre-existing failures on `12.1.x` (two AWS region-provider environment errors, one wallclock/timezone-boundary test) reproduce identically on base and are unrelated.

## Out of scope

The `FieldPartitioner` that builds the `<field>=<value>` encoded-partition path lives in `kafka-connect-storage-common` (a separate dependency); the audit's "Not covered" note calls for its own run.

### DIFF
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -429,9 +429,7 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
 
       configDef.define(
           AWS_ACCESS_KEY_ID_CONFIG,
-          // PASSWORD (not STRING) so AbstractConfig.logAll() renders it [hidden]: an AWS access
-          // key ID is a long-lived credential identifier and must not be logged in the clear.
-          Type.PASSWORD,
+          Type.STRING,
           AWS_ACCESS_KEY_ID_DEFAULT,
           Importance.HIGH,
           "The AWS access key ID used to authenticate personal AWS credentials such as IAM "
@@ -997,9 +995,7 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
   }
 
   public String awsAccessKeyId() {
-    // aws.access.key.id is now Type.PASSWORD; read it via getPassword so the parsed value
-    // resolves correctly (getString would ClassCastException on the Password object).
-    return getPassword(AWS_ACCESS_KEY_ID_CONFIG).value();
+    return getString(AWS_ACCESS_KEY_ID_CONFIG);
   }
 
   public Password awsSecretKeyId() {
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
@@ -429,7 +429,9 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
 
       configDef.define(
           AWS_ACCESS_KEY_ID_CONFIG,
-          Type.STRING,
+          // PASSWORD (not STRING) so AbstractConfig.logAll() renders it [hidden]: an AWS access
+          // key ID is a long-lived credential identifier and must not be logged in the clear.
+          Type.PASSWORD,
           AWS_ACCESS_KEY_ID_DEFAULT,
           Importance.HIGH,
           "The AWS access key ID used to authenticate personal AWS credentials such as IAM "
@@ -995,7 +997,9 @@ public class S3SinkConnectorConfig extends StorageSinkConnectorConfig {
   }
 
   public String awsAccessKeyId() {
-    return getString(AWS_ACCESS_KEY_ID_CONFIG);
+    // aws.access.key.id is now Type.PASSWORD; read it via getPassword so the parsed value
+    // resolves correctly (getString would ClassCastException on the Password object).
+    return getPassword(AWS_ACCESS_KEY_ID_CONFIG).value();
   }
 
   public Password awsSecretKeyId() {
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -85,7 +85,9 @@ public class AvroRecordWriterProvider extends RecordViewSetter
           public void write(SinkRecord record) throws IOException {
             if (!isWriterOpen()) {
               schema = recordView.getViewSchema(record, false);
-              log.info("Opening record writer for: {}", adjustedFilename);
+              // At DEBUG only: the filename embeds the encoded-partition path, which carries
+              // record field values under a field-based partitioner.
+              log.debug("Opening record writer for: {}", adjustedFilename);
               s3out = storage.create(adjustedFilename, true, AvroFormat.class);
               org.apache.avro.Schema avroSchema = avroData.fromConnectSchema(schema);
               writer.setCodec(CodecFactory.fromString(conf.getAvroCodec()));
--- a/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
+++ b/kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
@@ -85,10 +85,9 @@ public class AvroRecordWriterProvider extends RecordViewSetter
           public void write(SinkRecord record) throws IOException {
             if (!isWriterOpen()) {
               schema = recordView.getViewSchema(record, false);
-              // Log the record coordinates, not adjustedFilename: the filename embeds the
-              // encoded-partition path, which carries record field values under a field-based
-              // partitioner. It must not be logged at any level.
-              log.debug("Opening record writer for topic {} partition {} offset {}",
+              // Log Kafka coordinates, not adjustedFilename: its encoded-partition path can carry
+              // record field values under a field-based partitioner.
+              log.info("Opening record writer for topic {} partition {} offset {}",
                   record.topic(), record.kafkaPartition(), record.kafkaOffset());
               s3out = storage.create(adjustedFilename, true, AvroFormat.class);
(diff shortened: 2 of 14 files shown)


kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
```diff
@@ -85,7 +85,9 @@ private boolean isWriterOpen() {
           public void write(SinkRecord record) throws IOException {
             if (!isWriterOpen()) {
               schema = recordView.getViewSchema(record, false);
-              log.info("Opening record writer for: {}", adjustedFilename);
+              // At DEBUG only: the filename embeds the encoded-partition path, which carries
+              // record field values under a field-based partitioner.
+              log.debug("Opening record writer for: {}", adjustedFilename);
```
sumit-confluent: SME to confirm if removing adjustedFilename from the logger would make the debugging difficult).

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
```diff
@@ -77,7 +77,9 @@ public RecordWriter getRecordWriter(final S3SinkConnectorConfig conf, final Stri
           public void write(SinkRecord record) throws IOException {
             if (schema == null || writer == null) {
               schema = recordView.getViewSchema(record, true);
-              log.info("Opening record writer for: {}", adjustedFilename);
+              // At DEBUG only: the filename embeds the encoded-partition path, which carries
+              // record field values under a field-based partitioner.
+              log.debug("Opening record writer for: {}", adjustedFilename);
```
sumit-confluent: SME to confirm if removing adjustedFilename from the logger would make the debugging difficult.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
subhashiyer9: Yes. IMO, this will make debugging difficult.
This kind of logging has helped with multiple cases in the past.

subhashiyer9: If this applies only to FieldPartitioner, we can scope the change to FieldPartitioner alone?

subhashiyer9: I am also curious, if someone enables debugging, the same problem that is being solved here would surface again. Are we ok with that?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/parquet/ParquetRecordWriterProvider.java
subhashiyer9: Same comment as avro writer

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -112,7 +112,9 @@ public S3OutputStream(String key, S3SinkConnectorConfig conf, S3Client s3Client)
     this.position = 0L;
 
     this.enableConditionalWrites = conf.shouldEnableConditionalWrites();
-    log.info("Create S3OutputStream for bucket '{}' key '{}'", bucket, key);
+    // Log only the bucket, not the full S3 object key: under a field-based partitioner the key
+    // embeds record field values. The key remains available at DEBUG.
+    log.info("Create S3OutputStream for bucket '{}'", bucket);
```
subhashiyer9, line 117: Same comment as avro provider

kafka-connect-s3/src/main/java/io/confluent/connect/s3/S3SinkConnectorConfig.java
```diff
@@ -429,7 +429,9 @@ public static ConfigDef newConfigDef() {
 
       configDef.define(
           AWS_ACCESS_KEY_ID_CONFIG,
-          Type.STRING,
+          // PASSWORD (not STRING) so AbstractConfig.logAll() renders it [hidden]: an AWS access
+          // key ID is a long-lived credential identifier and must not be logged in the clear.
+          Type.PASSWORD,
```
subhashiyer9: Is this backwards compatible?
IMO, key is not considered a secret. Only the secret key is considered sensitive

subhashiyer9 (commented):
IMO, the changes limit the debuggability of the connector to track the lifecycle of a particular file

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
kumarsmConfluent: @subhashiyer9 , debug has been modified. Can you please re-look now?

subhashiyer9: @sumit-confluent @kumarsmConfluent 
Thoughts on this?
> If this applies only to FieldPartitioner, we can scope the change to FieldPartitioner alone?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
```diff
@@ -192,9 +193,8 @@ public void commit() throws IOException {
       }
     } catch (IOException e) {
       log.error(
-          "Multipart upload failed to complete for bucket '{}' key '{}'. Reason: {}",
+          "Multipart upload failed to complete for bucket '{}'. Reason: {}",
           bucket,
-          key,
```
subhashiyer9, line 197: can we log topic+partition+offset combo instead?

kafka-connect-s3/src/main/java/io/confluent/connect/s3/format/avro/AvroRecordWriterProvider.java
sumit-confluent: Rather than demoting to DEBUG (which would need partitioner-specific scoping so non-field partitioners do not lose the "opening writer" signal), the latest revision keeps this at INFO and logs the Kafka coordinates (topic/partition/offset) instead of the filename. That is safe for every partitioner — the encoded-partition path is never logged regardless of which partitioner is configured — so there is no need to special-case FieldPartitioner, and it keeps the file↔records correlation for debugging. Same change applied to ParquetRecordWriterProvider; JSON/ByteArray never logged the filename.

kafka-connect-s3/src/main/java/io/confluent/connect/s3/storage/S3OutputStream.java
sumit-confluent: topic/partition/offset are not available in `S3OutputStream` (it is constructed with just `key`/config/client, and one object spans many records across rotation, so there is no single offset triple here). I have logged the multipart `uploadId` as the correlator instead — it is opaque/non-sensitive, already logged elsewhere in this class (e.g. `abort()`), and is the natural handle for a failed multipart upload. The object key stays dropped since it embeds the encoded-partition path. Added a unit test (`testCommitFailureLogsUploadIdNotObjectKey`) asserting the failure ERROR carries the uploadId but not the object key. (pushed in 45b63e54)

subhashiyer9 (approved):

sumit-confluent:
Closing in favor of #964 — this PR is stuck on a pending `cla-status-check` that can't be resolved here. Same commits, recreated on a fresh branch.

PR #956, opened 2026-08-31
[Common BOM] Remove explicit easymock.version override, inherit from parent chain

nlou9:
## Summary
- Removes the explicit `<version>${easymock.version}</version>` override on `org.easymock:easymock` (plain `<dependencies>`, not `dependencyManagement`), letting it resolve from the parent chain instead.

## Not BOM-backed, and why that's OK here
Unlike most repos in this cleanup, this one's parent chain doesn't reach `confluent-common-bom`. `kafka-connect-storage-cloud` parents off `kafka-connect-storage-common-parent` (`12.0.12`), which itself pins `common [7.7.11, 7.7.12)` — a release that predates `confluent-common-bom`'s adoption entirely. Checked the latest published `kafka-connect-storage-common-parent` (`12.0.13`, one patch ahead of what this repo pins) — it's still on `common [7.7.11, 7.7.12)`, so this isn't a stale-branch issue on this repo's side (`master` is its only branch); the whole `kafka-connect-storage-common-parent` lineage hasn't been bumped to track `common`'s `8.x` line yet.

The removal is still a safe no-op — `common`'s own `dependencyManagement` at that version supplies the same `5.2.0` value either way, verified below.

## Verification
- `mvn -N validate` passes locally
- `mvn help:effective-pom` confirms `easymock` still resolves to `5.2.0`

## Tracking
Part of the `easymock.version` cleanup step in the [Common Inheritance Map — Cleanup Plan](https://claude.ai/code/artifact/854856e4-47a3-4455-9e22-326715a12c82) audit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #957, opened 2026-08-31
[Common BOM] Remove powermock.version property override

nlou9:
## Summary
- Removes the `powermock.version` override for `org.powermock:powermock-module-junit4` and `org.powermock:powermock-api-mockito2` (both plain `<dependency>` entries, so only the `<version>` lines needed to go — they inherit from the parent chain).
- `common`'s own `dependencyManagement` pins both at `${powermock.version}` = `2.0.9`, matching `confluent-common-bom`'s managed value exactly, so this is a no-op version change.
- Note: like the earlier `easymock.version` cleanup on this repo (#956), this is NOT actually BOM-backed end-to-end — `kafka-connect-storage-common-parent` (this repo's parent) is stuck pinning `common [7.7.11, 7.7.12)`, a structural gap unrelated to this change. The version still resolves correctly through `common`'s own `dependencyManagement`.
- Verified locally via `mvn -N validate` and `mvn help:effective-pom` (both resolve to `2.0.9`, unchanged).

## Tracking
Part of the `common` dependency-property cleanup effort: https://claude.ai/code/artifact/854856e4-47a3-4455-9e22-326715a12c82

## Test plan
- [x] `mvn -N validate` passes
- [x] `mvn help:effective-pom` confirms `powermock-module-junit4`/`powermock-api-mockito2` resolve to `2.0.9` (unchanged)
- [ ] CI green

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -425,7 +425,6 @@
         <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-module-junit4</artifactId>
-            <version>${powermock.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -436,7 +435,6 @@
         <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-api-mockito2</artifactId>
-            <version>${powermock.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>


nlou9:
Closing: this repo's resolved common version predates confluent-common-bom adoption (or the parent chain is structurally stuck pre-BOM), so this override doesn't actually migrate anything to BOM management -- it just trades one hardcoded-elsewhere value for another. Per updated policy, leaving repos alone when the resolved common version doesn't use confluent-common-bom for this dependency.

